PR #4, opened 2017-08-24
FIX: Adding prettytable as a dependency

hhslepicka:
Quick fix to pass the tests.


teddyrendahl:
@hhslepicka Can you also wipe out the way that `pedl` gets installed? I think I wrote this Travis file before I knew the cool way to install from Github via pip

teddyrendahl:
👍 👍

teddyrendahl:
Looks good merging in!

PR #10, opened 2017-09-20
ENH: Lightpath UI

teddyrendahl:
screen shot 2017-09-20 at 1 47 20 pm

* Built with PYDM which came with a whole host of dependencies
* Removed any need for `pcaspy` and `pedl`
* Played with the test simulation `lcls` a little bit to get the simulated GUI closer to the real thing

**ALSO**: This is being built with the branch https://github.com/slaclab/pydm/pull/81 from @hhslepicka , which solves a ton of issues for us. I'll leave it to you guys if we want to wait for it to be merged in and just use PyDM/master

**Not included in this PR**
* I want to build an easy to use launcher (give me JSON I give GUI) , but have to get rid of the circular dependencies of lightpath / pcds-devices would have on each other. This will be done shortly so keep your 👂 tuned

Closes #10


hhslepicka:
@mattgibbs and I are finishing the last adjustments needed for the PR on PyDM... it will be ready by Friday!

teddyrendahl:
Maybe we'll wait on this then ...

PR #17, opened 2017-10-01
MAINT: Improve branching logic

### LINKED ISSUE #15: ENH: Separate branch check from branching devices
When you reach a branch in the beamline, check upstream devices, not just the last one in the chain. This allows us to include devices immediately after mirrors that are technically on the same beamline, but don't inherently affect the branching.

teddyrendahl:
Closes #15 

### Branching Improvement
Prior to this commit, the branching logic in the lightpath required that the last device in every beamline be an optic. This however is not the case in actuality. In most cases involving mirrors, there are multiple devices allowed on the "shared" line after the optic but before the pointing has diverged into multiple beamlines. This update allows us to accomodate this within the lightpath as well as handle situations where pointing to different beamlines requires the coordination of multiple upstream optics i.e beam can not be delivered to MFX unless XRT M1H is removed AND XRT M2H is inserted. 

### Also
* Removed `pcaspy` from conda-recipe
* Some quick style fixes
* Add landscape cfg and badge

### DIFF
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,0 +1,16 @@
+doc-warnings : true
+test-warnings : false
+strictness: medium
+max-line-length : 80
+autodetect: true
+ignore-paths:
+   - docs
+   - lightpath/__init__.py
+ignore-patterns:
+    - versioneer.py
+    - lightpath/_version.py
+python-targets:
+    - 3
+pylint:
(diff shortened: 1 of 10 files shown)


teddyrendahl:
Nice catch @hhslepicka !

Issue #16, opened 2017-09-29
ENH: UI Improvement


teddyrendahl:
General list of Improvement Ideas
----------------------------------
- [ ]  Insert button for devices
- [x] Convert to a GridLayout for cleaner formatting
- [ ] Display path impediment at the top of the box. Focus on device, from click
- [ ] Frame entire row to indicate MPS status

teddyrendahl:
Removing `PyDMComboBox` request. This should be done by a https://github.com/pcdshub/typhon screen

PR #20, opened 2017-10-04
ENH: Show happi containers in UI

### LINKED ISSUE #19: FIX: Blank for improperly loaded devices
Device loading may fail for a variety of reasons; gateway issues, improper configuration e.t.c. In the current incarnation the `lightpath` ignores the device not drawing it at all. Instead the lightpath should draw some sort of blank indicating that device is there but we can not load it.

**Use case:**
IOC goes down with device blocking the beam, Python interface for device will not load correctly. Operator needs to know there is a device there ...

teddyrendahl:
Closes #19 and also part of #16 

screen shot 2017-10-04 at 4 32 26 pm

## Features
* Before, if a device failed to load, we would drop it from the lightpath. This could have some poor side affects if an IOC is down and the associated device is left in the beam the operators still need to be able to use the path as `map`. The GUI can now be passed a list of happi containers that will be considered "disconnected". The implementation was a bit of a pain as I had to find where each device fits along the path since I can't re-use the function internal to `LightController`
* The entire thing is now in a GridLayout. This will make it cleaner looking on different fonts and systems

## Questions
* Not in love with the blacked out lightpath for disconnected devices?
* After the switch to a GridLayout I lost the ability to put the entire row in a single frame. Is it worth the time to light the entire frame in case of an MPS fault, or am I just picking at scabs?


hhslepicka:
What about `purple` for the disconnected devices? For alarms we are using white... but I'm not sure how it will look like.
Would you mind making the UI so it does not require horizontal scrolling?
The UI looks great to me! I really liked it.

teddyrendahl:
@ZLLentz Yeah there is a way, but involves putting a frame around each cell and then specifically lighting up each side of the border. I can do it, just felt like it could be in a later PR. I'll add it to #16 

@hhslepicka The disconnected is purple? Or do you want all the text purple? Also I'm having trouble getting the UI to expand at all in horizontal. Maybe you and I can take a look.

hhslepicka:
@teddyrendahl 
The rectangle on the side of `FEE Valve 3` can be purple if you want to match the label `Disconnected`
PS: I'm really bad at color names and colors in general! 😞 
Sure... let's take a look at this expanding issue.

PR #27, opened 2017-10-31
MAINT: Use EPICS efficiently

### LINKED ISSUE #23: Minimize EPICS traffic
The `lightpath` is aggressive about checking device methods `inserted` and `removed`. This can be minimized by caching these somehow and instead trusting that we will receive updates via subscriptions. This will go a long way to making the UI more stable through the gateway e.t.c

### LINKED ISSUE #26: Alternate color for Unknown devices
`Unknown` devices should not be in red, it is too hard to differentiate them from the `Inserted`

teddyrendahl:
## Motivation
Closes #23 and closes #26 

EPICS calls are a resource! The`lightpath` shouldn't have blind trust that calls like `inserted` are implemented efficiently for each device. The UI should also be efficient using a single callback from the `BeamPath` object to configure all of the widgets instead of repeated calls. Here are a list of changes

## Change List
* `Unknown` is now displayed in yellow as not to be confused with `Inserted`. This is the only change to the UI
* All properties in `BeamPath` that make calls to `inserted` internally stash device positions. The idea is we should never be doing something like this which calls `device.inserted` for every device every time the property `inserted_devices` is requested.
```
if self.inserted_devices:
      return self.inserted_devices[0]
else:
      if self.inserted_devices > 5:
            return self.inserted_devices[:5]
```
Thinks weren't this obvious, but since calls like `impediment`, `blocking_devices` all made calls to `device.inserted` we were far from efficient
* `BeamPath` will only run subscriptions if a device is changed upstream of the `impediment`. We do not care if the 12th device changes state if the path is blocked at the first
* `BeamPath` uses the devices `SUB_STATE` event_type per https://github.com/slaclab/pcds-devices/pull/89
* `BeamPath` does not subscribe to child devices unless someone else subscribes to it
* Each `LightWidget` was subscribed to `BeamPath`. Each callback referenced `incident_devices`, this mean that when the `BeamPath` updated we called `incident_devices` * number of devices. This was a HUGE waste of time. Now ALL `BeamPath` callbacks are done once, changes are then propagated down to update the appropriate widgets. The individual widget still subscribes to its own device to reflect changes in device state.
* Cleaned up a few logging messages

### DIFF
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -46,11 +46,11 @@ class LightController:
         #Iterate through creating complete paths 
         for bp in sorted(self.beamlines.values(),
                          key = lambda x : x.path[0].z):
-            logger.debug("Assembling complete beamline %s ...", bp.name)
+            logger.info("Assembling beamline %s ...", bp.name)
 
             #Grab branches off the beamline
             for branch in bp.branches:
-                logger.info("Found branches onto beamlines %s from %s",
+                logger.debug("Found branches onto beamlines %s from %s",
                              ', '.join(branch.branches), branch.name)
                 for dest in branch.branches:
                     if dest != bp.name:
--- a/lightpath/path.py
+++ b/lightpath/path.py
@@ -464,9 +464,13 @@ class BeamPath(OphydObject):
             #Subscribe to all child devices
             for dev in self.devices:
                 #Add callback here!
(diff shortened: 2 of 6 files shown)


teddyrendahl:
A few commits after launching with real devices. Ready for merge!

PR #29, opened 2017-11-04
MAINT: name is now a required keyword-only argument.

ZLLentz:
API update for newest Ophyd compatibility in tests. I tried to change as few lines as possible.


teddyrendahl:
Seems good @ZLLentz . Travis is still building with `0.8.0` specifically. Should we change it?

ZLLentz:
@teddyrendahl I was planning to swap them all over to 1.0.0 specifically

ZLLentz:
@teddyrendahl the updated tags in the lightsource2-tag channel are 0.8.0 and 0.11.0, so we'll keep those for now

PR #39, opened 2017-11-23
CI: Use fixed happi version

teddyrendahl:
## Description

There maybe be non backwards compatible changes in the happi directory soon. I locked the version number down until these are done.

## Motivation and Context

Trying to avoid accidentally breaking things

## How Has This Been Tested?

Tested via Travis. Using the tagged happi build should be no different than using the current master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
   #Grab all dependencies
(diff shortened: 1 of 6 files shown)


teddyrendahl:
Fixed to use single `=`

teddyrendahl:
Fixed `QApplication` tear down bug. Ready for merge

PR #45, opened 2017-12-01
BLD: Build with the ui file

ZLLentz:
## Description

Add the ui file to the `MANIFEST.in` file, set `include_package_data=True`
## Motivation and Context

This means we can distribute lightpath as a conda package and it will work.

## How Has This Been Tested?

python setup.py build creates the correct file structure


teddyrendahl:
Thanks for looking at this!

PR #46, opened 2018-02-01
MAINT: Update to latest pcdsdevices, happi

teddyrendahl:
## Description

This pull request makes `lightpath` work with `happi > 0.5.x` and `pcdsdevices > 0.3.x`. The major changes that affected the lightpath were:

1. Metadata is no longer tacked on to the device. All references to `device.z` had to be changed to `device.md.z`
2. `MPS` devices are no longer associated with devices. They will be in the `happi` database as a separate object
3. All device loading should be done in happi.

For the actual `BeamPath` code, I deprecated all the references to the MPS system. In the future, we will just create a small object `MPSSummary` that manages the entire MPS system for a beamline. These will also be displayed in their own panel in the purpose of the GUI.

The most frustrating change is that the UI used to treat `happi.Device` objects and `pcdsdevices` objects the same. Now that the latter has metadata contained in the `md` attribute, that code does not work. For now I have just decided to remove them from the main lightpath GUI. I think the answer in the end is to have a separate panel where disconnected devices are displayed.

## Motivation and Context

Should now be importable in newer releases of `pcds-envs`

In the end, this PR is a step back in terms of feature set but nothing that was heavily used was taken away. It can be used in the new `hutch-python` environment. The rest will be addressed during the UI re-release.

## How Has This Been Tested?

Updated tests to match new library expectations. Also created a `BeamPath` in the current `mfxpython` environment to do a live check.

### DIFF
--- a/lightpath/tests/conftest.py
+++ b/lightpath/tests/conftest.py
@@ -3,6 +3,7 @@
 ############
 import logging
 from enum import Enum
+from types import SimpleNamespace
 
 ###############
 # Third Party #
@@ -25,7 +26,7 @@ from lightpath import BeamPath
 def pytest_addoption(parser):
     parser.addoption("--log", action="store", default="INFO",
                      help="Set the level of the log")
-    parser.addoption("--logfile", action="store", default=None,
+    parser.addoption("--logfile", action="store", default='log',
                      help="Write the log output to specified file path")
 
 #Create a fixture to automatically instantiate logging setup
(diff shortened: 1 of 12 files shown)


lightpath/tests/conftest.py
```diff
@@ -25,7 +26,7 @@
 def pytest_addoption(parser):
     parser.addoption("--log", action="store", default="INFO",
                      help="Set the level of the log")
-    parser.addoption("--logfile", action="store", default=None,
+    parser.addoption("--logfile", action="store", default='log',
```
ZLLentz: Are you finding these command line arguments useful?

teddyrendahl: No no at all. Should be refactored to the `/logs` folder setup. That commit got snuck in.

ZLLentz: Ah, okay. I was just fishing for more workflow improvement potential for other modules.

PR #51, opened 2018-02-14
API: Controller

### LINKED ISSUE #18: Periscope Pathing
After talking to Teddy about this, I went through the code and thought about this issue a lot more... I think it deserves a write-up. Apologies in advance for the wall of text:

The addition of the periscope line causes us issues compared to the others because the path is no longer a nice tree structure, it becomes a more general directed graph. This is a quick brain dump about this problem and potential solutions.

Issues:
- Destination is not enough to identify a beamline when there are 2 or more equivalent paths to the end.
- In the current schema, `BeamPath` objects must end if they don’t have a branching device.

These two issues combined mean that we must split the XCS lines into a ‘Periscope’ line that ends before the LODCM and an ‘XCS’ line that goes all the way through, including the inside of the LODCM (e.g. make it look like a tree). This is ok for skywalker but not for a more general use case in XCS, and we may have this problem again with some of the proposed FEE branching schemes in lcls2.

So we must do the following:
- Modify the `beamlines` `dict` in `LightController` to allow two paths to exist with the same destination, without compromising usability.
- Create a mechanism for `BeamPath`s to continue at the start of other `BeamPath`s without a physical branching device.
- Bonus: create a mechanism for `BeamPath`s to continue in the middle of other `BeamPath`s (Not required, just makes definitions cleaner. May be useful in lcls2)

Implementation idea:
- Add two node devices to explicitly mark graph nodes when they can’t be determined implicitly. These would be stripped from `BeamPath` objects at the end of `__init__`, but would direct the way these paths are built:
  - `ContinueNode`: place at the end of a `BeamPath` that should continue onto the beginning of another `BeamPath`. This would simply implement the branches property and specify a list of one element: the destination line.
  - `JoinNode`: place at the beginning of a `BeamPath` that should accept inputs from multiple `BeamPath`s. This should implement a new property, `sources`, that is a `dict` mapping of source `BeamPath` name to a suffix for the combined line.

So, with current hardware, we’d have an XCS_Periscope `BeamPath` and an XCS_MONO `BeamPath` that each end in a `ContinueNode`, where `node.branches == [‘XCS’]`. The XCS line would begin with a `JoinNode` right before LODCM Cr2, where `node.sources == dict(XCS_Periscope=’P’, XCS_MONO=’M’)`. The `LightController` sees the branch to XCS on each line, so it calls `xcspath.join`. It then checks for the `JoinNode` and renames the `BeamPath` accordingly, stashing it into the `beamlines` `dict`. Eventually it reaches the XCS line itself, sees the `JoinNode`, deletes XCS from beamlines (it's already been duplicated/split if it has a `JoinNode`), and goes through the branches block for the new `BeamPath`s, XCSP and XCSM, instead (in case there was another branch point). At the end, all `ContinueNode` and `JoinNode` objects are stripped out, and we're left with two valid `BeamPath`s to the XCS interaction point, XCSP and XCSM. (Example names of course)

Let me know what you think, we need to properly handle these double paths at some point. I think the concept of node objects will be useful for things like continuing the HXD line into CXI as well.

### LINKED ISSUE #32: Selective loading
Only load the devices necessary for the requested path. This should avoid gateway issues, decrease loading time ...

### LINKED ISSUE #47: Hard code path configuration
## Expected Behavior

The information on the LCLS beamlines is simple and static enough that it should be kept in a configuration in the file. This simplifies the management of the controller object itself.

teddyrendahl:
### API
* Instead of passing already configured devices into `LightController` you pass a `happi.Client`. The `LightController` is then responsible for selecting the correct devices and instantiating them. 
* `LightApp` takes a `LightController` object instead of a list of devices.

### Enhancements
* The `LightController` has an optional `endstations` keyword that will only load a list of hutches beampaths. 

### Maintenance
* Prior to this PR, the `LightController` used a tangled web of code to stitch together beamlines based on the device it found. This is now hard coded in `lightpath/config`. There is a `beamlines` dictionary where each endstation is a key with a nested dictionary describing the complete path. You can just leave the smaller dictionary blank indicating you want to include the entire beamline, or you can optionally specify a `start` and/or `end` to select part of a section of beamline.

### Fixes
* Fixes a bug where we looked for `branches` property on the `md` sub-attribute rather than the device. The tests were mistakenly updated in a prior PR as well. These have both been rectified. 

## Motivation and Context

Closes #32 
Closes #47 

Do you think this closes #18? This new configuration will create two paths `PBT` and `XCS`. This greatly improves the situation as the `PBT` path will include devices in the XCS hutch itself.

## How Has This Been Tested?

Minimal changes to tests. Updated to new API. 

Updated docstrings

### DIFF
--- a/lightpath/config.py
+++ b/lightpath/config.py
@@ -1,0 +1,13 @@
+"""
+Configuration for LCLS beamlines
+"""
+post_xrtm2h = 817.2
+xpp_lodcm = 781.2
+
+beamlines = {'XPP': {'HXD': {'end': xpp_lodcm}},
+             'XCS': {'HXD': {}},
+             'PBT': {'HXD': {'end': post_xrtm2h},
+                     'XCS': {}},
+             'MFX': {'HXD': {'end': post_xrtm2h}},
+             'CXI': {'HXD': {}},
+             'MEC': {'MEC': {'end': post_xrtm2h}}}
--- a/lightpath/config.py
+++ b/lightpath/config.py
@@ -10,4 +10,4 @@ beamlines = {'XPP': {'HXD': {'end': xpp_lodcm}},
                      'XCS': {}},
              'MFX': {'HXD': {'end': post_xrtm2h}},
              'CXI': {'HXD': {}},
-             'MEC': {'MEC': {'end': post_xrtm2h}}}
+             'MEC': {'HXD': {'end': post_xrtm2h}}}
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -6,20 +6,15 @@ contsructed as a :class:`.BeamPath` object. This includes not only devices on
 the upstream beamline but all of the beamlines before it. For example, the MEC
 beampath will include devices in both the FEE and the XRT. The
 :class:`.LightController` handles this logic as well as a basic overview of
-where the beam is and what the state of the MPS system is currently. 
+where the beam is and what the state of the MPS system is currently.
 """
-####################
-# Standard Library #
-####################
+import math
 import logging
-####################
-#    Third Party   #
-####################
 
-####################
-#     Package      #
-####################
+from happi.loader import from_container
+
 from .path import BeamPath
+from .config import beamlines
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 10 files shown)


lightpath/config.py
```diff
@@ -0,0 +1,13 @@
...
+post_xrtm2h = 817.2
+xpp_lodcm = 781.2
+
+beamlines = {'XPP': {'HXD': {'end': xpp_lodcm}},
+             'XCS': {'HXD': {}},
+             'PBT': {'HXD': {'end': post_xrtm2h},
+                     'XCS': {}},
+             'MFX': {'HXD': {'end': post_xrtm2h}},
+             'CXI': {'HXD': {}},
+             'MEC': {'HXD': {'end': post_xrtm2h}}}
```
ZLLentz: 👍

lightpath/controller.py
```diff
@@ -6,20 +6,15 @@
...
-####################
+import math
 import logging
-####################
-#    Third Party   #
-####################
 
-####################
-#     Package      #
-####################
```
ZLLentz: 👍

PR #60, opened 2018-05-09
ENH: Launch Script

### LINKED ISSUE #56: Startup script
## Expected Behavior

We should be able to have a `lightpath` executable that takes a `happi` configuration file and displays a GUI.

## Current Behavior

This is boilerplate code that has to be repeated in launch scripts.

```python
from PyQt5.QtGui import QApplication
from lightpath import Controller
from lightpath.ui import LightApp

app = QApplication()
light = LightApp(Controller('path/to/db.json'))
light.exec_()
```

teddyrendahl:
## Description

Launch script to show GUI for arbitrary `happi` database.

```bash
lightpath --db PATH_TO_DB
```

I hardcoded in the path where it currently lives as a default. I though about importing this from `hutch_python` but adding a dependency for just a single path seems gratuitous. 

## Motivation and Context

Closes #56 

## How Has This Been Tested?

* Added test for script packaging
* Tested with real database

### DIFF
--- a/lightpath/tests/test_gui.py
+++ b/lightpath/tests/test_gui.py
@@ -1,3 +1,5 @@
+from distutils.spawn import find_executable
+
 from lightpath.ui import LightApp
 from lightpath.controller import LightController
 
@@ -23,3 +25,8 @@ def test_beampath_controls(lcls_client):
     assert lightapp.rows[0].device.inserted
     lightapp.transmission_adjusted(50)
     assert lightapp.path.minimum_transmission == 0.5
+
+
+def test_lightpath_launch_script():
+    # Check that the executable was installed
+    assert find_executable('lightpath')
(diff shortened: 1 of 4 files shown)


lightpath/tests/test_gui.py
```diff
@@ -23,3 +25,8 @@ def test_beampath_controls(lcls_client):
     assert lightapp.rows[0].device.inserted
     lightapp.transmission_adjusted(50)
     assert lightapp.path.minimum_transmission == 0.5
+
+
+def test_lightpath_launch_script():
+    # Check that the executable was installed
+    assert find_executable('lightpath')
```
ZLLentz: didn't know this utility existed

teddyrendahl:
Looks like a bug unless you can interpret this ...
screen shot 2018-05-08 at 7 59 54 pm

ZLLentz:
Actually, it looks like (somehow) your repr was being called in a test (?) and now it is not. No idea.

PR #62, opened 2018-05-11
MAINT: Handle faulty devices

### LINKED ISSUE #61: Overview BeamPath methods should be robust against exceptions
## Expected Behavior

When calling functions that give you an overview of the entire `BeamPath` object such as `show_devices`, we should be prepared for devices to raise Exceptions. One disconnected device should not prevent an operator from seeing the overview. These devices should instead report an `Error` state.

## Current Behavior

```python
In [2]: mfx_beampath.show_devices()
/reg/g/pcds/pyps/conda/py36/envs/pcds-0.6.0/lib/python3.6/site-packages/epics/ca.py:1330: UserWarning: ca.get('STPR:XRT1:1:SH2_PPSSUM') timed out after 1.00 seconds.
...
ValueError: None is not a valid state for sh2. Valid state names are: ['Unknown', 'IN', 'OUT'], and their corresponding values are [0, 1, 2].
```
This was noticed today while the FEE gateway was overloaded.

Full ELog post -> https://pswww.slac.stanford.edu/apps/portal/index.php?exper_id=794&app=elog:search&params=message:381604

## Possible Solution

```try / except```

### PR DESCRIPTION
**Pull request by teddyrendahl**
## Description

The `lightpath` wasn't suitably protected against faulty devices. Many of the methods shamelessly called `device.inserted` e.t.c not accepting that these usually involve Channel Access calls that are liable to fail. Now, we only ask for device information in `find_device_state`. This returns a `DeviceState` enum that describes the state. This helps limit the number of Channel Access calls and in turn uniformly handle when these things fail.

## Motivation and Context

Closes #61 

## How Has This Been Tested?

Added a test for a broken device

### DIFF
--- a/lightpath/path.py
+++ b/lightpath/path.py
@@ -178,16 +178,15 @@ class BeamPath(OphydObject):
             # Find branching devices and store
             # They will be marked as blocking by downstream devices
             try:
+                dev_state = find_device_state(device)
                 if device in self.branches:
                     last_branches.append(device)
-
                 # Find inserted devices
-                elif device.inserted and (device.transmission <
-                                          self.minimum_transmission):
-                    block.append(device)
-                # Find unknown devices
-                elif not device.removed and not device.inserted:
-                    block.append(device)
+                elif dev_state == DeviceState.Inserted:
+                    # Ignore devices with low enough transmssion
+                    trans = getattr(device, 'transmission', 1)
+                    if trans < self.minimum_transmission:
+                        block.append(device)
             except Exception as exc:
                 logger.error('Unable to determine state of %s', device.name)
                 logger.error(exc)
@@ -206,7 +205,8 @@ class BeamPath(OphydObject):
         inserted but have more transmission than :attr:`.minimum_transmission`
         """
         # Find device information
-        inserted = [d for d in self.path if d.inserted]
+        inserted = [d for d in self.path
+                    if find_device_state(d) == DeviceState.Inserted]
         impediment = self.impediment
         # No blocking devices, all inserted devices incident
         if not impediment:
@@ -224,7 +224,7 @@ class BeamPath(OphydObject):
             File to writable
         """
         # Initialize Table
-        pt = PrettyTable(['Name', 'Prefix', 'Position', 'Beamline', 'Removed'])
+        pt = PrettyTable(['Name', 'Prefix', 'Position', 'Beamline', 'State'])
         # Adjust Table settings
         pt.align = 'r'
         pt.align['Name'] = 'l'
@@ -233,7 +233,7 @@ class BeamPath(OphydObject):
         # Add info
         for d in self.path:
             pt.add_row([d.name, d.prefix, d.md.z,
-                        d.md.beamline, str(d.removed)])
+                        d.md.beamline, find_device_state(d).name])
         # Show table
         print(pt, file=file)
 
@@ -291,7 +291,9 @@ class BeamPath(OphydObject):
         logger.info('Removing devices along the beampath ...')
         status = [device.remove(timeout=timeout)
                   for device in target_devices
-                  if not device.removed and hasattr(device, 'remove')]
+                  if find_device_state(device) in (DeviceState.Inserted,
+                                                   DeviceState.Unknown)
+                  and hasattr(device, 'remove')]
         # Wait parameters
         if wait:
             logger.info('Waiting for all devices to be '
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lightpath/path.py [hunk 1]
```diff
@@ -27,6 +28,43 @@
...
+    except Exception as exc:
+        logger.exception("Unable to determine device state for %r", device)
+        return DeviceState.Faulted
+    # Check state consistency and return proper Enum
+    if _in and not _out:
+        return DeviceState.Inserted
+    elif _out and not _in:
+        return DeviceState.Removed
+    else:
+        return DeviceState.Unknown
```
ZLLentz: This method is :+1: :+1:, great simple solution! Puts all of the error checking in one place and introduces a faulted state for the bad situations and broken code.

Does this mean we should have devices raise exceptions when they know they are in a faulted state?

lightpath/path.py
```diff
@@ -253,7 +288,9 @@ def clear(self, wait=False, timeout=None,
         logger.info('Removing devices along the beampath ...')
         status = [device.remove(timeout=timeout)
                   for device in target_devices
-                  if not device.removed and hasattr(device, 'remove')]
+                  if find_device_state(device) in (DeviceState.Inserted,
+                                                   DeviceState.Unknown)
+                  and hasattr(device, 'remove')]
```
ZLLentz: I don't like four-line list comprehensions for lack of readability, but that's more a matter of personal taste. This seems to do the job.

lightpath/tests/test_path.py
```diff
@@ -199,14 +223,14 @@ def test_complex_branching(lcls):
 
 known_table = """\
 +-------+--------+----------+----------+---------+
-| Name  | Prefix | Position | Beamline | Removed |
+| Name  | Prefix | Position | Beamline |   State |
```
ZLLentz: :+1:

ZLLentz:
btw the assigned tab still has me there after the review, so mission accomplished

lightpath/path.py
teddyrendahl: Hmmm. Can you give me an example of device we would want to do this? In my mind, `Faulted` would be reserved for a disconnection issue or something beyond just not being in any of the designated states. (In the `clear` method I opted to not even to try and call `.remove` on faulted devices)

ZLLentz: Ah, so a device that knows it's in a bad state either reports "in" and "out", or neither "in" nor "out"? Maybe I'm tripping over semantics for things like the `Attenuator` which has a status PV that shows faulted in some scenarios.

teddyrendahl: Yeah maybe we should think about this, current logic is:
not in or not out -> "Unknown"
in and out -> "Unknown"
unable to report -> "Faulted"

That second one might want to be "Faulted" . My original thought was that we wanted to distinguish connection issues from just things being in the wrong place. Helpful for operators to be like this is "Unknown" I'm going to try and move this, and "Faulted" this IOC is down we need to call someone

ZLLentz: So our discussion was:
not connected -> disconnected
in and not out -> inserted
out and not in -> removed
not in and not out -> unknown
exception raised -> error
seem reasonable?

teddyrendahl:
Alright. Per discussion here and offilne, I launched the whole gambit of states.  `DeviceState` now is very specific about what we know about the device.
```
    Removed:
        Device is removed from the beamline.

    Inserted:
        Device is inserted into the beamline. This may or may not prevent beam
        from reaching downstream devices.

    Unknown:
        Device is reporting neither an inserted or removed state.

    Inconsistent:
        The device is reporting that is both inserted and removed.

    Disconnected:
        We were unable to determine the state of the device because one or more
        of the relevant signals was not available.

    Error:
        Catch-all state for any errors the device reported when asked for its
        state that were not simply a failure to communicate with signals
```

lightpath/path.py
```diff
@@ -53,14 +85,27 @@ def find_device_state(device):
...
+                     device.name, _in, _out)
     except Exception as exc:
+        # Catch DisconnectionError
+        if isinstance(exc, DisconnectedError):
+            logger.warning("Unable to connect to %r", device)
+            return DeviceState.Disconnected
+        # Check if this was an error with an EPICS connection
         logger.exception("Unable to determine device state for %r", device)
-        return DeviceState.Faulted
+        return DeviceState.Error
```
ZLLentz: Should probably add `TimeoutError` too in case this happens in a signal `get` call, which will try to wait for connection.

The canonical way to check for different error types is:
```
except (DisconnectedError, TimeoutError):
    <stuff>
except Exception:
    <stuff>
```
Rather than calling `isinstance` yourself
```

ZLLentz:
:+1:

PR #71, opened 2018-07-17
UI: Row Change

teddyrendahl:
## Description

Load each row from a preconfigured `.ui` file instead of creating the entire widget in code. This leads to a better separation of "ui" code and actual "lightpath" logic. The overall PR actually adds lines of code, but if you look by making the UI file I was able to delete quite a few lines of actual Python code. The new UI file also makes the following changes. 
* Buttons are stacked vertically
* Prefix and alias are stacked vertically
* Each device is separated by a horizontal line
* Improvements to layout specification to handle different screen sizes

This was going to part of a large PR that included `Typhon` screens but I've gone back and forth too many times on that implementation and chose just to push this snippet as its own entity.

screen shot 2018-07-16 at 6 33 04 pm

### DIFF
--- a/lightpath/ui/device.ui
+++ b/lightpath/ui/device.ui
@@ -1,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Form</class>
+ <widget class="QWidget" name="Form">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>620</width>
+    <height>94</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <property name="spacing">
+    <number>5</number>
+   </property>
+   <property name="leftMargin">
+    <number>0</number>
+   </property>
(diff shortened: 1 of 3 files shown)


ZLLentz:
I'm a fan of the reorganization to make and edit the rows as .ui files. I'm still not fond of making the action buttons laid out vertically because it makes the application less intuitive to use from a muscle memory standpoint. It's much better to have an insert column and a remove column because the scrolling is vertical.

As laid out here, the insert and remove buttons are actually changing places as we scroll down the screen.

teddyrendahl:
screen shot 2018-07-17 at 3 44 06 pm

## Changes
* I placed the buttons to be next to each other horizontally
* In order to make the spacing consistent now the `insert` and `remove` buttons will **always** be there, we just make sure to enable and disable them. This helps everything stay nice and consistent.
* Did some maintenance work on moving more items to the device widget where it belongs.

ZLLentz:
Looks better, I like the enable/disable solution. Not crazy about the wasted space, I wonder if it's worth having a compact mode... I'd like to be able to see a large swath of the beamline at once to scan for the blocking device. Maybe a separate issue.

How does this fit in with setting an attenuation on the screen? Or is that all going to be through typhon?

PR #72, opened 2018-07-18
UI: Beamline

teddyrendahl:
## Description

Part of the transformation to making the UI look more like the old system. We now create a mock "drawing" of the beamline that lights up according to the device state and whether the beam is being transmitted or not.

screen shot 2018-07-18 at 2 44 47 pm

### DIFF
--- a/lightpath/tests/test_widgets.py
+++ b/lightpath/tests/test_widgets.py
@@ -3,8 +3,7 @@ from unittest.mock import Mock
 import pytest
 
 import lightpath.ui
-from lightpath.ui.widgets import state_colors
-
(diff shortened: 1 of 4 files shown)


hhslepicka:
Nice design... quick question... is there a case where the name on top is different than the name under it? E.g. FEE Valve 1 and (Fee Valve 1)

teddyrendahl:
@hhslepicka Should have shared what the version on live devices looks like. That will be the PV prefix for real devices. I don't put it there in the simulation so I don't get them confused 😉

hhslepicka:
Awesome! Thank you for the reply @teddyrendahl ! ;)

PR #76, opened 2018-07-19
ENH: Filter Options

### LINKED ISSUE #67: Ability to hide removed devices from table
## Expected Behavior

In the most common mode of operation beamline operators should only be concerned with devices that are not removed from the beamline. This helps with the current information overload of the current screen.

### LINKED ISSUE #73: Hide widgets instead of removing them
## Expected Behavior

Disconnecting and reconnecting widgets is expensive. We should only delete them if we are confident we won't want them again in the near future. This is largely for when you show / hide the upstream devices along a beamline

teddyrendahl:
## Description

A bunch of helpful navigation widget options. You can now:
* Force scroll to an arbitrary device
* Force scroll to the most upstream blocking device
* Hide devices that are removed from the beam
* Hide devices based on their class

I also implemented the `upstream` filter in the more general structure now. This means we **only** disconnect / reconnect to widgets when we switch beamlines. Avoid loading the entire beampath whenever possible!

## Motivation and Context

Closes #73 
Closes #67 

## How Has This Been Tested?

Added tests for all new functions

screen shot 2018-07-19 at 4 26 01 pm

### DIFF
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -8,6 +8,7 @@ import os.path
 
 import numpy as np
 import pcdsdevices.device_types as dtypes
+from pcdsdevices.valve import PPSStopper
 from pydm import Display
 from pydm.PyQt.QtCore import pyqtSlot, Qt
 from pydm.PyQt.QtGui import QHBoxLayout, QGridLayout, QCheckBox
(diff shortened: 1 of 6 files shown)


lightpath/ui/gui.py
```diff
@@ -1,14 +1,18 @@
 """
 Full Application for Lightpath
 """
+from functools import partial
 import logging
 import threading
 import os.path
 
+import numpy as np
+from pcdsdevices.device_types import *  # noqa
```
ZLLentz: `*` imports are the devil, but we'll let it slide... though it's odd to have style rules that get dumped when it's convenient.

teddyrendahl: Yeah. I wanted to avoid two giant lists imports and then accepted classes, this felt cleaner. You disagree?

ZLLentz: It's slightly messier, but I would prefer something like:
```
import pcdsdevices.device_types as dtypes
shown_types = [dtypes.Attenuator, dtypes.IPM, dtypes.XFLS, ...]
```
So it's clear where the classes in `shown_types` come from.

ZLLentz: The fact that you need to `# noqa` it is a nudge that it's making your code less clear

teddyrendahl: Good point. Didn't think to shorten. Will change

teddyrendahl:
Fixed the imports, added a test. Nothing too controversial. Will merge if no complaints.

ZLLentz:
:+1:

PR #79, opened 2018-07-31
ENH: Add version command to lightpath

teddyrendahl:
## Description

Needed to stretch my legs with a real easy PR today. Added the `--version` option to the `lightpath` startup script. Useful to tell where and which `lightpath` module you'll be using when you enter the command.

## Motivation and Context

@hhslepicka

### DIFF
--- a/bin/lightpath
+++ b/bin/lightpath
@@ -5,8 +5,8 @@ import coloredlogs
 
 import happi
 import pydm
+
 import lightpath
-from lightpath import LightController
 from lightpath.ui import LightApp
 
 DEVICE_CONFIG = '/reg/g/pcds/pyps/apps/hutch-python/device_config/db.json'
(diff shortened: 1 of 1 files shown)


bin/lightpath
```diff
@@ -5,6 +5,7 @@ import coloredlogs
 
 import happi
 import pydm
+import lightpath
```
ZLLentz: Your `lightpath` imports are starting to look a bit messy up here

ZLLentz:
```
import lightpath
from lightpath import LightController
from lightpath.ui import LightApp
```

teddyrendahl:
Better @ZLLentz?

ZLLentz:
:+1: It wasn't a huge deal as-is, but these things can get out of hand quickly

PR #86, opened 2018-09-13
FIX: Multiple Hidden Conditions

### LINKED ISSUE #81: Removed devices / upstream devices fight each other
We need to make sure that if you multiple filters on, clearing one will still respect the others.

For example:

```
1. Hide upstream devices, these are all hidden.
2. Hide removed devices
3. Un-hide removed devices. -> This third call can not cause upstream devices to be visible.
```

teddyrendahl:
## Description

There was a bug where multiple hidden conditions would override each other. This behavior was rather unintuitive. For a widget to be visible it should have to meet all the selected criteria not just one. This PR consolidates the hidden logic into a single `filter` function which is then triggered by all the buttons.

## Motivation and Context

Closes #81 

## How Has This Been Tested?

* Modified test suite for new filtering functions
* Added a check for multiple filtering functions being active simultaneously

### DIFF
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -53,6 +53,7 @@ class LightApp(Display):
         self.light = controller
         self.path = None
         self.detail_screen = None
+        self.device_buttons = dict()
         self._lock = threading.Lock()
         # Create empty layout
         self.lightLayout = QHBoxLayout()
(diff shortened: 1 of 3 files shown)


lightpath/ui/gui.py
```diff
@@ -273,28 +280,24 @@ def focus_on_device(self, name=None):
...
+            hidden_removed = (not self.remove_check.isChecked()
+                              and row[0].last_state == DeviceState.Removed)
+            # Hide if upstream
+            beamline = self.selected_beamline()
+            hidden_upstream = (not self.upstream_check.isChecked()
+                               and device.md.beamline != beamline)
+            # Hide device if any of the criteria are met
+            row[0].setHidden(hidden_device_type
+                             or hidden_removed
+                             or hidden_upstream)
```
ZLLentz: quick reminder that if this gets any bigger you may favor the `any` built-in instead of using an `or` chain

PR #90, opened 2018-10-09
BLD: Pin typhon to <=2.1

teddyrendahl:


teddyrendahl:
Will do. Putting this as a reminder myself largely!

PR #94, opened 2019-03-28
FIX: Include all ui files in package

teddyrendahl:
Classic bug! Unfortunately this one snuck into a release. Not packaging `.ui` files properly...


teddyrendahl:
Rebased to pin `typhon` version instead of patching.

PR #95, opened 2019-06-26
CI: Install xvfb as a Travis service

teddyrendahl:


.travis.yml
```diff
@@ -15,23 +19,19 @@ matrix:
         env:
            - PCDS_CHANNEL=pcds-dev
       - python: 3.7
-        dist: xenial
         sudo: true
         env:
            - PCDS_CHANNEL=pcds-tag
       - python: 3.7
-        dist: xenial
         sudo: true
```
klauer: Remove all `sudo: true`, along with the call to `apt-get update` below

PR #100, opened 2020-03-18
MNT: Compat with Latest Happi

ZLLentz:
## Description

Happi enforces that `name` must be a valid Python identifier now. I did a sloppy find/replace so that the tests pass again.

### DIFF
--- a/lightpath/tests/path.json
+++ b/lightpath/tests/path.json
@@ -1,6 +1,6 @@
 {
-    "FEE Valve 1": {
-        "_id": "FEE Valve 1",
+    "fee_valve1": {
+        "_id": "fee_valve1",
         "active": true,
         "args": [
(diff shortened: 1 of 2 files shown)


ZLLentz:
I think there is something funky going on with loading devices with invalid fields, while loading devices with missing fields seems to be okay...

hhslepicka:
@ZLLentz I think as long as you don't try to set the fields with an invalid value it is okay.
`save` would get the missing required fields but at instantiation it will barf if the value does not meet the required format.

ZLLentz:
@hhslepicka then maybe the name field is special (because it is doubling as the id_ field?). Loading a database with an invalid name does not work.

ZLLentz:
Unless some local saving is happening in this testing suite

PR #103, opened 2020-03-23
REF: lightpath script -> entry point

klauer:
Also pep8 and isort fixes

Untested locally

### DIFF
--- a/lightpath/launcher.py
+++ b/lightpath/launcher.py
@@ -13,7 +13,8 @@ DEVICE_CONFIG = '/reg/g/pcds/pyps/apps/hutch-python/device_config/db.json'
 
 logger = logging.getLogger('lightpath')
 
-def main(db, hutches):
+
+def lightpath(db, hutches):
     """
     Open the lightpath user interface for a configuration file
 
@@ -34,7 +35,7 @@ def main(db, hutches):
     app.exec_()
 
 
-if __name__ == '__main__':
+def main():
     # Create ArgumentParser
     parser = argparse.ArgumentParser(description='Launch the Lightpath UI')
     parser.add_argument('--db', dest='db', type=str,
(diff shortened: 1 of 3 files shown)


lightpath/launcher.py
```diff
@@ -4,16 +4,16 @@
 import coloredlogs
 
 import happi
-import pydm
-
 import lightpath
+import pydm
```
ZLLentz: This is now the wrong sort order. The original order was correct. I'm not very pleased with isort based on this and other instances...

lightpath/tests/conftest.py
```diff
@@ -1,16 +1,17 @@
...
-from ophyd.utils import DisconnectedError
 
+import happi
 import lightpath
 from lightpath import BeamPath
+from ophyd import Component as Cpt
+from ophyd import Device, Kind
+from ophyd.signal import AttributeSignal
+from ophyd.status import DeviceStatus
+from ophyd.utils import DisconnectedError
```
ZLLentz: This is again the wrong sort order, unless I misunderstand the rules? I'm going to stop and look them up again...

ZLLentz: Yes, this is definitely wrong. Stuff like `ophyd` are third-party imports, not local imports. https://www.python.org/dev/peps/pep-0008/#imports

lightpath/tests/test_widgets.py
```diff
@@ -1,11 +1,11 @@
...
 
-from ophyd import Device
 import pytest
 
 import lightpath.ui
-from lightpath.ui.widgets import (state_colors, to_stylesheet_color,
-                                  symbol_for_device)
+from lightpath.ui.widgets import (state_colors, symbol_for_device,
+                                  to_stylesheet_color)
+from ophyd import Device
```
ZLLentz: Another botched import order

ZLLentz:
I stopped commenting on all of them, but `isort` seems to get it very wrong because it cannot properly differentiate between a `third-party` library and a `local` library.

ZLLentz:
Code looks good though

PR #115, opened 2020-07-22
MNT: Fixes for new deployment

### LINKED ISSUE #112: Lightpath opens extra PyDM Main Window
## Expected Behavior

Should not spawn an empty main window

## Current Behavior

Extra main window loaded
image

## Possible Solution

Read PyDM docs and update lightpath code

## Steps to Reproduce (for bugs)

1. Open lightpath UI

## Context

If it annoys me, it will annoy users

## Your Environment

Latest

### LINKED ISSUE #114: Lightpath Ignores Happi's Lightpath Flag
## Expected Behavior

Expect `active=True`, `lightpath=True` -> use in the path

## Current Behavior

`active=True`, `lightpath=False` -> is also used in the path

## Possible Solution

extend the happi load calls

ZLLentz:
## Description

Accumulated small fixes PR for lightpath redeployment
closes #112 
closes #114 
partially addresses #116 probably

### DIFF
--- a/lightpath/ui/widgets.py
+++ b/lightpath/ui/widgets.py
@@ -6,7 +6,7 @@ import os.path
 
 from pydm import Display
 from ophyd import Kind
-from qtpy.QtCore import Signal as pyqtSignal, Qt
+from qtpy.QtCore import Signal, Qt
 from qtpy.QtGui import QBrush, QColor, QFont
 from qtpy.QtWidgets import QLabel
 import qtawesome as qta
@@ -120,7 +120,7 @@ class LightRow(InactiveRow):
     parent : QObject, optional
     """
     MAX_HINTS = 2
-    device_updated = pyqtSignal()
+    device_updated = Signal()
 
     def __init__(self, device, parent=None):
         super().__init__(device, parent=parent)
@@ -226,7 +226,7 @@ class DeviceWidget(QLabel):
     device: ophyd.Device
         Object that will have a drawing created for it.
     """
-    clicked = pyqtSignal()
+    clicked = Signal()
 
     def __init__(self, device, parent=None):
         super().__init__(parent=parent)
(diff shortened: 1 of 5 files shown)


lightpath/ui/widgets.py
```diff
@@ -120,13 +120,16 @@ class LightRow(InactiveRow):
     parent : QObject, optional
     """
     MAX_HINTS = 2
+    device_updated = pyqtSignal()
```
hhslepicka, line 123: maybe switch it to use `qtpy` and `Signal` instead of `pyqtSignal` ?

lightpath/ui/widgets.py
```diff
@@ -187,7 +193,8 @@ def update_state(self, *args, **kwargs):
         self.state_label.setText(self.last_state.name)
         color = state_colors[self.last_state.value]
         style_color = to_stylesheet_color(color)
-        self.state_label.setStyleSheet("QLabel {color: %s}" % style_color)
+        style_sheet = "QLabel {color: %s}" % style_color
```
hhslepicka, line 196:
```suggestion
        style_sheet = f"QLabel {{color: {style_color} }}"
```

lightpath/ui/widgets.py
ZLLentz: https://github.com/pcdshub/lightpath/blob/415410dfa810af825337580c437409b51b6a416d/lightpath/ui/widgets.py#L9

ZLLentz: I guess this is how Teddy updated to qtpy. Should I take the opportunity to make it clearer by removing the as import?

hhslepicka: It seems like a good idea!

lightpath/ui/widgets.py
ZLLentz: I'm way too scared of the nested `}` here to put this in

PR #119, opened 2021-01-12
MNT: noarch, shared travis config

ZLLentz:
## Description

- make recipe noarch
- use shared travis config
- enable python>=3.8

## Motivation and Context

- related to https://github.com/pcdshub/pcds-envs/pull/158#issuecomment-758859276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,95 +1,25 @@
-language: python
-dist: xenial
-sudo: false
-services:
-    - xvfb
+version: ~> 1.0
 
 env:
   global:
     - OFFICIAL_REPO="pcdshub/lightpath"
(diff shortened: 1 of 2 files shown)


ZLLentz:
The build output here is very confusing. Probably something in dev-requirements.txt breaks the install, but only after switching to strict channel priority...

klauer:
Hmm 😕  

Think you'd be up to reconsidering adding our miscellaneous, LCLS-specific dependencies to conda-forge? I have an inkling that if all the packages were there, we might be able to avoid some of these headaches.

ZLLentz:
Oh for sure, it would probably solve a lot of related problems. For some of them I'd prefer to just eat them into pcdsdevices, though given the lack of effort on my part to actually do that it seems like a meaningless point.

ZLLentz:
I think in this case the travis would still explode, but the full env build would be closer to passing

ZLLentz:
for some reason, the built package here (and confirmed locally) is only installable on python 3.9... I notice the build string has a py39 in it despite being noarch. I'm a bit puzzled, I need to read some conda docs and compare recipes.

klauer:
I've seen some of our other CI implicitly picking Python 3.9 confusingly - I can't recall details for this to be an actually useful anecdote unfortunately. 😦

ZLLentz:
There was an error in the recipe, I was not correctly designating noarch. Hopefully this works out better.

Issue #120, opened 2021-02-17
Installed Lightpath Segfaults


ZLLentz:
## Current Behavior

Throws this exception then segfaults:
```
NotADirectoryError: [Errno 20] Not a directory: '/cds/home/z/zlentz/miniconda3/envs/path36/lib/python3.6/site-packages/lightpath-0.6.7-py3.9.egg/lightpath/ui/lightapp.ui'
```
Tests pass in CI because they use the lightpath.ui from the cloned repo, but the packaging is messed up.

I have no idea what changed or why this would stop working, but we are definitely doing the conda packaging differently in `lightpath` vs the other repositories.

## Possible Solution

- Edit the conda recipe to standardize things
- Go conda-forge
- Edit the CI to catch this sort of case in the tests

## Steps to Reproduce (for bugs)

1. `conda install lightpath`
2. `lightpath`

## Your Environment

pcds-4.0.0
See https://github.com/pcdshub/pcds-envs/issues/169

ZLLentz:
`conda-recipe/build.sh` is the suspicious file here for me, but I still don't know why everything would suddenly break.

ZLLentz:
closed by #121

PR #121, opened 2021-02-17
BLD: rework setup.py with console entrypoint, package_data

klauer:
(I see no difference with this locally, may just leave this unmerged...)

Related #120


ZLLentz:
May be worth merging and seeing if installed from the dev channel like this works... otherwise I think we need to find a way to make the tests show failure

klauer:
Yeah, that's a fair point - worth a shot, at least.

ZLLentz (approved):
Let's try it

ZLLentz:
This worked

PR #133, opened 2022-06-29
WIP/ENH: use graphs for path finding, and apply new Lightpath interface

tangkong:
# Still very much a work in progress

## Description
This is a doozy, but I think it's time to start writing some of this down.  

- applies the new Lightpath interface from https://github.com/pcdshub/pcdsdevices/pull/1028/ to the `BeamPath` and `BeamController` objects
- uses this interface to represent the facility as a directed graph
  - uses simple path finding algorithms to find routes to a given beamline
  - sets out methods for handling multiple paths to the same destination
- Adds and updates tests accordingly 
  - Adds a new `path.json` file that roughly reflects the current LCLS beamlines 

## Motivation and Context
Lightpath in its current incarnation cannot represent the LCLS facility.  This representation is hopefully more robust, and allows for new devices and beam paths to be added.

## How Has This Been Tested?
A best effort has been made to keep the old tests working, but new tests will be soon to come.

This PR, to start.  We'll see how much of my brain-dump style notes make it in here.  Perhaps a confluence page is worthwhile

### DIFF
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -11,6 +11,8 @@ where the beam is and what the state of the MPS system is currently.
 import logging
 import math
 
+import networkx as nx
+
 from .config import beamlines
 from .path import BeamPath
 
@@ -42,14 +44,51 @@ class LightController:
         self.client = client
         self.containers = list()
         self.beamlines = dict()
+        self.graph = nx.DiGraph()
+        self.sources = set()
+
+        # initialize graph -> self.graph
+        self.load_facility()
+
         endstations = endstations or beamlines.keys()
-        # Find the requisite beamlines to reach our endstation
+        # # Find the requisite beamlines to reach our endstation
         for beamline in endstations:
             self.load_beamline(beamline)
 
+    def load_facility(self):
+        """
+        Load the facility from the provided happi client.
+
+        The facility is represented by a directed graph, with
+        SearchResults as nodes.
+        """
+        results = self.client.search_range(key='z', start=0.0, end=math.inf,
+                                           active=True, lightpath=True)
+        # gather devices by branch
+        branch_dict = {}
+        for res in results:
+            for branch_set in (res.metadata.get('input_branches', []),
+                               res.metadata.get('output_branches', [])):
+                for branch in branch_set:
+                    if branch not in branch_dict:
+                        branch_dict[branch] = set()
+                    branch_dict[branch].add(res)
+
+        # Construct subgraphs and merge
+        subgraphs = []
+        for branch_name, branch_devs in branch_dict.items():
+            subgraph = BeamPath.make_graph(branch_devs,
+                                           sources=beamlines['sources'],
+                                           branch_name=branch_name)
+            self.sources.update((n for n in subgraph if 'source' in n))
+            subgraphs.append(subgraph)
+
+        self.graph = nx.compose_all(subgraphs)
+
     def load_beamline(self, endstation):
         """
-        Load a beamline from the provided happi client
+        Load a beamline given the facility graph.  Find path from source
+        to endstation with reasonable transmission
 
         Parameters
         ----------
@@ -61,43 +100,35 @@ class LightController:
         path: BeamPath
         """
         try:
-            path = beamlines[endstation]
-            path[endstation] = dict()
+            end_branches = beamlines[endstation]
         except KeyError:
             logger.warning("Unable to find %s as a configured endstation, "
                            "assuming this is an independent path", endstation)
-            path = {endstation: {}}
+            # TODO: Look at all branches?  What to do here
+            return
 
-        # Load the devices specified in the configuration
-        devices = list()
+        paths = list()
+        for branch in end_branches:
+            # Find the paths from each source to the desired line
+            for src in self.sources:
+                if nx.has_path(self.graph, src, branch):
+                    found_paths = nx.all_simple_paths(self.graph, source=src,
+                                                      target=branch)
+                    paths.extend(found_paths)
+                else:
+                    logger.debug(f'No path between {src} and {branch}')
 
-        for line, info in path.items():
-            # Find the happi containers for this section of beamlines
-            start = info.get('start', 0.0)
-            end = info.get('end', math.inf)
-            logger.debug("Searching for devices on line %s between %s and %s",
-                         line, start, end)
-            results = self.client.search_range(key='z', start=start, end=end,
-                                               beamline=line, active=True,
-                                               lightpath=True)
-            # Ensure we actually found valid devices
-            if not results:
-                logger.error("No valid beamline devices found for %s", line)
-                continue
-            # Load all the devices we found
-            logger.debug("Found %s devices along %s", len(results), line)
-            for result in results:
-                try:
-                    dev = result.get()
-                    devices.append(dev)
-                except Exception:
-                    logger.exception("Failure loading %s ...", result["name"])
-                    self.containers.append(result.device)
-        # Create the beamline from the loaded devices
-        bp = BeamPath(*devices, name=line)
-        self.beamlines[line] = bp
-        # Set as attribute for easy access
-        setattr(self, bp.name.replace(' ', '_').lower(), bp)
+        self.beamlines[endstation] = []
+        for path in paths:
+            subgraph = self.graph.subgraph(path)
+            devices = [node[1]['dev'] for node in subgraph.nodes.data()
+                       if node[1]['dev'] is not None]
+            bp = BeamPath(*devices, name=endstation)
+            self.beamlines[endstation].append(bp)
+
+        for bp in self.beamlines[endstation]:
+            # may be a list of beampaths, for destinations with many paths
+            setattr(self, bp.name.replace(' ', '_').lower(), bp)
         return bp
 
     @property
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -241,18 +241,62 @@ class LightController:
                 )
                 continue
             nodes.append((res.metadata['name'], {'dev': dev}))
-        # add end point node
-        nodes.append((branch_name, {'dev': None}))
+
+        # construct edges
+        edges = []
+        skipped_right = []
+        skipped_left = []
+        last_on_branch = 0
+        edata = {'weight': 0.0, 'branch': branch_name}
+        # Need to properly handle devices that either:
+        # only have the current branch in their input (skipped_right)
+        # only have the current branch in their output (skipped_left)
+        for i in range(len(nodes)):
+            curr_dev = nodes[i][1]['dev']
+            if (branch_name in curr_dev.input_branches and
+                    branch_name in curr_dev.output_branches):
+                if last_on_branch != i:
+                    # attach skipped devices
+                    for ri in skipped_right:
+                        edges.append((nodes[last_on_branch][0],
+                                     nodes[ri][0], edata))
+                    skipped_right = []
+
+                    for li in skipped_left:
+                        edges.append((nodes[li][0], nodes[i][0], edata))
+                    skipped_left = []
+
+                    # make edge between last on branch and this on dev
+                    # luckily duplicate edges are ignored
+                    edges.append((nodes[last_on_branch][0],
+                                 nodes[i][0], edata))
+                # update last_on_branch
+                last_on_branch = i
+
+            try:
+                next_dev = nodes[i+1][1]['dev']
+            except IndexError:
+                # we are at the end, skip steps that look ahead
+                continue
+
+            if set(curr_dev.output_branches) & set(next_dev.input_branches):
+                # base case, make edge as normal
+                edges.append((nodes[i][0], nodes[i+1][0], edata))
+            elif (branch_name not in curr_dev.input_branches):
+                # skip this, attach it to next
+                skipped_left.append(i)
+            elif (branch_name not in curr_dev.output_branches):
+                skipped_right.append(i)
+
         # add sources
         if branch_name in sources:
             nodes.insert(0, (f'source_{branch_name}', {'dev': None}))
+            edges.append((nodes[0][0], nodes[1][0], edata))
+        # add end point
+        nodes.append((branch_name, {'dev': None}))
+        edges.append((nodes[-2][0], nodes[-1][0], edata))
 
         graph.add_nodes_from(nodes)
-
-        # construct edges
-        edges = [(nodes[i][0], nodes[i+1][0],
-                 {'weight': 0.0, 'branch': branch_name})
-                 for i in range(len(nodes)-1)]
         graph.add_edges_from(edges)
 
         graph.name = str(branch_name)
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -13,7 +13,7 @@ import math
 
 import networkx as nx
 
-from .config import beamlines
+from .config import beamlines, sources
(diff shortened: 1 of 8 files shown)


lightpath/happi/containers.py
```diff
@@ -0,0 +1,32 @@
+"""
+Define subclasses of Happi Items for use with Lightpath
+"""
+import re
+from copy import deepcopy
+
+from happi.item import EntryInfo, OphydItem
+
+
+class LightpathItem(OphydItem):
```
klauer, line 10: I take it this intentionally does not inherit from `LCLSItem` to make it accessible to those not at the LCLS?

lightpath/path.py
```diff
@@ -171,86 +176,81 @@ def __init__(self, *devices, name=None):
                             'attribute {}'.format(e))
 
     @property
-    def branches(self):
-        """
-        Branching devices along the path
-        """
-        return [d for d in self.devices if getattr(d, 'branches', False)]
+    def branches(self) -> List[Device]:
```
klauer, line 179: Since we're breaking API and everything here, it may be time for some renames

`branches` -> `branching_devices`?

lightpath/path.py
```diff
@@ -171,86 +176,81 @@ def __init__(self, *devices, name=None):
...
-            elif dev_state == DeviceState.Inserted:
+            curr_state = find_device_state(device)
+            try:
+                curr_status = device.get_lightpath_status()
+            except Exception as e:
+                logger.debug(f'lightpath status unknown: {e}')
+                curr_status = None
+
+            # short circuit if statuses are in error
+            if ((curr_state is DeviceState.Error) or
```
klauer, line 219: Can be a bit less rigorous here - `curr_state in (DeviceState.Error, DeviceState.Unknown)` would be just fine.

May also be worthwhile to switch `DeviceState` to either `IntEnum` or a string enum depending on how it's used - bare enum is a bit inconvenient after all

lightpath/path.py
```diff
@@ -171,86 +176,81 @@ def __init__(self, *devices, name=None):
...
-            # Find branching devices and store
-            # They will be marked as blocking by downstream devices
-            dev_state = find_device_state(device)
-            if device in self.branches:
-                last_branches.append(device)
-            # Find inserted devices
-            elif dev_state == DeviceState.Inserted:
+            curr_state = find_device_state(device)
+            try:
+                curr_status = device.get_lightpath_status()
```
klauer, line 213: It looks like `find_device_state` calls `get_lightpath_status`, so we're calling it twice to get the same information (however unlikely, it could also change between calls)

That helper probably should just return both the status and the state

klauer (commented):
Making some real progress - looking good! 👍

lightpath/happi/containers.py
tangkong: Yep!  Though the only test cases I have are based off the LCLS facility.

lightpath/controller.py
```diff
@@ -42,115 +49,402 @@ def __init__(self, client, endstations=None):
...
         self.beamlines = dict()
+        self.graph = nx.DiGraph()
+        self.sources = set()
+
+        # initialize graph -> self.graph
+        self.load_facility()
+
         endstations = endstations or beamlines.keys()
-        # Find the requisite beamlines to reach our endstation
+        # # Find the requisite beamlines to reach our endstation
```
ZLLentz, line 59:
```suggestion
        # Find the requisite beamlines to reach our endstation
```

lightpath/controller.py
```diff
@@ -42,115 +49,402 @@ def __init__(self, client, endstations=None):
...
+        The facility is represented by a directed graph, with
+        devices as nodes.  Edge weights are initialized as 0, and
+        labeled with their branch.
+
+        The facility graph is created by combining subgraphs that
+        each contain all the devices on a given branch.
         """
-        Load a beamline from the provided happi client
+        results = self.client.search_range(key='z', start=0.0, end=math.inf,
+                                           active=True, lightpath=True)
```
ZLLentz, line 75: The main thing this has over a search for just `active=True, lightpath=True` is the enforcement that `z` exists and is non-negative, right? Some day we may consider making the "position" key configurable... Today is not that day.

lightpath/controller.py
```diff
@@ -42,115 +49,402 @@ def __init__(self, client, endstations=None):
         self.client = client
         self.containers = list()
         self.beamlines = dict()
+        self.graph = nx.DiGraph()
```
ZLLentz, line 52: This empty graph is overwritten on line 95 when we load the facility. Should this be a type annotation instead?

lightpath/controller.py
```diff
@@ -42,115 +49,402 @@ def __init__(self, client, endstations=None):
...
+        Returns
+        -------
+        float
+            z position of impediment
+        """
+        return getattr(path.impediment, 'md.z', math.inf)
+
+    def active_path(self, dest: str) -> BeamPath:
+        """
+        Return the most active path to the requested endstation
```
ZLLentz, line 160: Is this intended to help for cases like XCS where there are 2 routes?

lightpath/happi/containers.py
ZLLentz: Should be :+1: to eventually inherit this in pcdsdevices

ZLLentz (commented):
I didn't try to run the app like this (branch juggling) but this seems pretty solid. I'm excited and hopeful that the formalization of the path as a directed graph + the other changes will allow us to fix issues we have regarding pathing in the RIX/TMO area among others.

lightpath/controller.py
tangkong: Yep, Those are the conditions for a device to be valid in lightpath's concept of a facility.  I thought about trying to make the "position" key more configurable, but the concept should be reasonably extensible to other light sources (synchrotrons just define z as distance ccw from injection or something?)

lightpath/controller.py
tangkong: That was the idea.  The tests for this method primarily use XCS routes.  (and they pass I promise)

tangkong:
These tests are fine locally, but there's some mismatch between the happi container and path.json showing up on travis... 😢

tangkong:
The happi error turned out to be due to the fact that the pip build does not actually install the package, meaning entry points don't get picked up.  The test database therefore cannot find the `lightpath.happi.containers.LightpathItem` container.  This was fixed by adding `-e ./` to `PIP_EXTRAS`. 

This led to a failure to tear down QThreads in the pip test suite, which I could not fix.  PIP failures have been allowed, and the python builds limited to 3.9 (latest) to allow for more modern type hinting.  

With all that, the tests pass 🎉 Not the most complete victory but a finished one

lightpath/path.py
```diff
@@ -171,125 +185,114 @@ def __init__(self, *devices, name=None):
...
-            elif dev_state == DeviceState.Inserted:
+            curr_state, curr_status = find_device_state(device)
+
+            # short circuit if statuses are in error
+            if curr_state in (DeviceState.Error, DeviceState.Unknown):
+                block.append(device)
+
+            # check to make sure input and output branches match
+            # e.g. mirror not pointing to current device
+            elif (prev_device and prev_status and
```
klauer, line 228:
```suggestion
            elif (prev_device is not None and prev_status is not None and
```

lightpath/path.py
```diff
@@ -489,10 +513,15 @@ def _device_moved(self, *args, obj=None, **kwargs):
         else:
             block = math.inf
         # If device is upstream of impediment
-        if obj and obj.md.z <= block:
+        if obj and obj.parent.md.z <= block:
```
klauer, line 516:
```suggestion
        if obj is not None and obj.parent.md.z <= block:
```
Side note: the only object which has metadata is the root `Device`, so this is implicitly a component on the device with a lightpath mixin. That's an OK assumption to make here.

klauer (approved):
Almost all of this looks great to me. I added a bunch of comments and suggestions which may be useful to you - or feel free to ignore/resolve straightaway (without even an emoji response, if you prefer!).

I loaded this up with the provided test database and it works as advertised.

Mostly unrelated to your changes, I/we should make issues out of these:
1. It would be nice if the GUI respected the normal happi client configuration mechanism (via `HAPPI_CFG`) and didn't circumvent it by specifying `DEVICE_CONFIG` in `lightpath.__main__`
2. It'd be nice if `python -m lightpath` worked (it might require a bit of restructuring due to how the entrypoint is set up, annoyingly - don't look at git blame there 👀 )
3. `lightpath_summary` in typhos never shows connected; something is weird there. It should also be marked as "kind=omitted" for all devices (other PR).
4. I see a bit of icon clipping in the summary at the top of the GUI:
image

lightpath/happi/containers.py
tangkong: The conflict shows itself as:
``` 
(pcds-5.4.1)roberttk@psbuild-rhel7-01:~/devrepos/pcdsdevices(enh_lp_mixin +)$ happi container-registry
File "/cds/home/r/roberttk/devrepos/happi/happi/containers.py", line 208, in load
    self._safe_add(entry_name, var)
  File "/cds/home/r/roberttk/devrepos/happi/happi/containers.py", line 173, in _safe_add
    raise RuntimeError(f"Duplicated entry found. Keys: {key} "
RuntimeError: Duplicated entry found. Keys: pcdsdevices.happi.containers.LightpathItem and lightpath.happi.containers.LightpathItem point to same class: <class 'lightpath.happi.containers.LightpathItem'>
```

When you import `from lightpath.happi.containers import LightpathItem` in `pcdsdevices.happi.containers`

It seems the container gets collected once in lightpath and again in pcdsdevices under different keys, since they're being collected from different entrypoints.  I can add an issue for it

klauer: I see now - yeah, this is a problem. Needs more thought

lightpath/controller.py
```diff
@@ -42,115 +50,403 @@ def __init__(self, client, endstations=None):
...
+        -------
+        Dict[NodeName, List[NodeName]]
+            A mapping from source node names to path of nodes
+
+        Raises
+        ------
+        PathError
+            If a single, valid path cannot be determined
+        """
+        sources = [n for n in self.graph.nodes if 'source' in n]
```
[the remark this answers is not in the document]
tangkong: Turns out we populate the facility sources during `load_facility` but I just ... forgot.  😞

lightpath/controller.py
```diff
@@ -42,115 +50,403 @@ def __init__(self, client, endstations=None):
...
+
+        paths: Dict[NodeName, List] = {k: [] for k in sources}
+
+        for src, path in paths.items():
+            successors = list(self.graph.successors(src))
+            # skip to node after source node
+            curr = successors[0]
```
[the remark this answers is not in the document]
tangkong: I believe there's some protection against that particular case since we only add to sources if a subgraph with that branch name exists.  Still good to add an exception catch here though

lightpath/controller.py
```diff
@@ -42,115 +50,403 @@ def __init__(self, client, endstations=None):
...
+            while successors:
```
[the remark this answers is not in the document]
tangkong: Yep, that's my intention here.  I suppose it's probably worth explaining myself in the docstrings more for future readers.

lightpath/controller.py
```diff
@@ -42,115 +50,403 @@ def __init__(self, client, endstations=None):
...
+        graph = nx.DiGraph()
+        result_list = list(branch_devs)
+        result_list.sort(key=lambda x: x.metadata['z'])
+        # label nodes with device name, store ophyd device
+        nodes = []
+        for res in result_list:
+            try:
+                dev = res.get()
+            except Exception:
+                # TODO: be better about specific exceptions
```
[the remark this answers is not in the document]
tangkong: This one deserves some more thought for sure.  The knockon effects are pretty significant, since this would break the facility graph and path-finding methods.  

The dummy device that we'd have to make would also have to give a best effort guess at the input/output branches.  I might punt this to another PR, but will be necessary

lightpath/__init__.py
```diff
@@ -2,6 +2,6 @@
 
 from . import _version
 from .controller import LightController
-from .path import BeamPath
+from .path import BeamPath, LightpathState  # noqa F401
```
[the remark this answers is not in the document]
klauer: I completely agree about avoiding star imports, but as far as I can tell `__all__` is the only other recognized mechanism for "public API". So if it's not part of the public API through one of the above 2 mechanisms, it probably just shouldn't be there.

Based on the statement "`LightpathState` is a piece of the device interface" I do think it seems reasonable to keep it as public.

lightpath/path.py
```diff
@@ -171,125 +185,114 @@ def __init__(self, *devices, name=None):
...
+            # check inserted
+            elif curr_state is DeviceState.Inserted:
                 # Ignore devices with low enough transmssion
-                trans = getattr(device, 'transmission', 1)
-                if trans < self.minimum_transmission:
+                if curr_status.transmission < self.minimum_transmission:
                     block.append(device)
             # Find unknown and faulted devices
-            elif dev_state != DeviceState.Removed:
+            elif curr_state is not DeviceState.Removed:
```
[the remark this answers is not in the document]
tangkong: This should be caught by the above condition, no?  I suppose we could make this more explicitly an `else` statement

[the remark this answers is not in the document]
klauer: I was quick to jump on this - I wanted to make sure I saw every `DeviceState` referenced once and got distracted with the actual logic. 

Just in case, there should be an `else` clause, too, that is considered the same as "errored".

lightpath/controller.py
```diff
@@ -42,115 +50,403 @@ def __init__(self, client, endstations=None):
...
+        edges: List[Tuple[NodeName, NodeName, Dict[str, Any]]] = []
+        skipped_right: List[int] = []
+        skipped_left: List[int] = []
+        last_on_branch = 0
+        # weight not used currently, but may be for path finding algos
+        edata = {'weight': 0.0, 'branch': branch_name}
+        # Need to properly handle devices that either:
+        # only have the current branch in their input (skipped_right)
+        # only have the current branch in their output (skipped_left)
+        for i in range(len(nodes)):
```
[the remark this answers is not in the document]
tangkong: This one was a tougher one for me to figure out as well, I'll try to clarify in the comments.  

This logic is complicated because it's trying to handle "dangling" nodes.  These nodes either only have the current branch as an input or only as an output.  These nodes are distinct from branching devices, and I had to make sure only the correct edges were constructed.  

I couldn't figure out a concise way to describe the logic, but maybe something more concise would be more confusing to read through

lightpath/path.py
```diff
@@ -67,46 +80,50 @@ class DeviceState(enum.Enum):
     Error = 5
 
 
-def find_device_state(device):
+def find_device_state(device: Device) -> Tuple[DeviceState, LightpathState]:
     """
     Report the state of a device
 
     The device must implement ``.inserted`` and ``removed``.
```
ZLLentz, line 87: Stale docstring, maybe?

tangkong: I've gone blind looking at this code I guess

ZLLentz (approved):
This looks solid to me, we'll need to battle-test it a bit but it looks good

tangkong:
Thanks for the thorough reviews, I'm going to merge this with the intent to tackle outstanding issues in separate PR's.

PR #141, opened 2022-07-21
ENH: find happi db from config, add sim option

### LINKED ISSUE #134: ENH: make GUI respect happi client config
## Expected Behavior
It would be nice if the GUI respected the normal happi client configuration mechanism (via HAPPI_CFG) and didn't circumvent it by specifying DEVICE_CONFIG in lightpath.__main__

## Context
_Originally posted by @klauer in https://github.com/pcdshub/lightpath/pull/133#pullrequestreview-1036074786_

## Your Environment
pcds-5.4.1

tangkong:
## Description
- Removes hard coded `DEVICE_CONFIG`
- loads a happi client from config by default
- adds an option to run lightpath with the sim database

## Motivation and Context
closes #134 
Also sneaks in this sim option which might be useful later

## How Has This Been Tested?
tests run locally.  interactive testing shows the database getting the right path, but currently the database is not up-to-date.

CLI documentation updated.

### DIFF
--- a/lightpath/__main__.py
+++ b/lightpath/__main__.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+from pathlib import Path
 
 import coloredlogs
 import happi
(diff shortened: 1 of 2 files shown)


lightpath/__main__.py
```diff
@@ -60,8 +63,12 @@ def entrypoint():
     else:
         hutches = None
 
+    # set up simulated lightpath if requested
+    if args.sim:
```
klauer, line 67: `if args.sim and not args.db` (in case I want my own simulation database?)

klauer (approved):
LGTM; the `--sim` mode fix is a nice touch.

Unsolicited thought, as per usual: it'd be especially neat if we could use our regular database in "sim" mode and have happi automatically give us fake devices back somehow...

lightpath/__main__.py
tangkong: There's nothing special about the sim version of the app, so if you want your own test database you can just pass it in through --db.  If we do implement the "sim" mode with fake devices, this would make sense though.

klauer: Oh, _right_...

ZLLentz (approved):
This looks like a clear :+1: :+1: to me

Issue #137, opened 2022-07-12
Properly handle beamlines that end in the middle of a branch


tangkong:
## Expected Behavior
Beamlines that end in the middle of a branch should only include devices up to their beamline.  

## Current Behavior
Currently we expect beamlines to be located at the end of a branch.  This works in most cases, but some beamlines (XPP) live in the middle of a branch, so their active paths end up including a variety of irrelevant devices.

## Possible Solution
This will probably require a slightly different way of specifying where a beamline sits.  Perhaps we also allow a device name to designate a beamline.  

## Steps to Reproduce (for bugs)
initialize the test database (`path.json`) and look at the active paths for `XPP`:

Note how the first beam path includes all the `L0` devices, even though XPP should really end before `xpp_lodcm`
```python
In [2]: lc.beamlines['XPP'][0].show_devices()
+-----------+-------------+-----------+----------------+--------------------+---------+
| Name      | Prefix      |  Position | Input Branches |    Output Branches |   State |
+-----------+-------------+-----------+----------------+--------------------+---------+
| im1l0     | IM1L0:XTES  | 699.50000 |         ['L0'] |             ['L0'] | Removed |
| sl1l0     | SL1L0:POWER | 705.86000 |         ['L0'] |             ['L0'] | Removed |
| im2l0     | IM2L0:XTES  | 715.50000 |         ['L0'] |             ['L0'] | Removed |
| mr1l0     | MR1L0       | 740.00000 |         ['L0'] |       ['L0', 'L1'] | Removed |
| im3l0     | IM3L0:XTES  | 749.50000 |         ['L0'] |             ['L0'] | Removed |
| im4l0     | IM4L0:XTES  | 755.50000 |         ['L0'] |             ['L0'] | Removed |
| sl2l0     | SL2L0:POWER | 755.86000 |         ['L0'] |             ['L0'] | Removed |
| xpp_lodcm | XPP:LODCM   | 781.10000 |         ['L0'] |       ['L0', 'L2'] | Removed |
| mr1l3     | MR1L3       | 814.71600 |         ['L0'] |       ['L0', 'L3'] | Removed |
| sl3l0     | SL3L0:POWER | 815.86000 |         ['L0'] |             ['L0'] | Removed |
| mr1l4     | MR1L4       | 817.11600 |         ['L0'] | ['L0', 'L4', 'L5'] | Removed |
[...]

PR #142, opened 2022-07-27
Follow-up bug fixes, gui details

tangkong:
## Description
Fixes a slew of issues/errors found while running the gui on the updated happi db.  Individual problems will be documented below.
<details> <summary>Detailed problem documentation</summary>

### 1. EPICS Context unset on teardown
- **Problem**: on teardown, an immense amount of traceback initiated by EPICS CA contexts being unset.
- **Explanation**: On shutdown, nothing is done to gracefully complete/close any in-progress callbacks.  
- **Fix**: Hooked the closeEvent method to run a clean-up method that destroys all the `lightpath_summary` signals before closing.  (This might be a bit heavy handed, but it worked)

### 2. GUI subscriptions should subscribe to `lightpath_summary` signal
- **Problem**: widget subscriptions still subscribed to the whole device with SUB_STATE.  This resulted in failures of the widget to subscribe to some devices, due to SUB_STATE being missing.
- **Fix**: subscribe to `lightpath_summary` directly in the LightRow widget.  Remember to initialize the `SummarySignal` by calling `lightpath_summary.get()` after subscribing

### 2.1. Getting stuck waiting on a Lock forever while calling `lightpath_summary.get()`
- **Problem**: After changing the subscription method in LightRow, GUI initialization would hang.  
- **Explanation**: This was tracked down to an issue with locks colliding.  Signal callbacks subscribed by the GUI will run while the `lightpath_summary._lock` is claimed/locked.  These callbacks also had a chance to try and subscribe widgets to devices, which in turn called `lightpath_summary.get()` and attempted to grab `lightpath_summary._lock`.  
- ~~**easy fix**: keep track of if `lightpath_summary.get()` has been called, and only call it once~~
- **better fix**: don't call `lightpath_summary.get()` at all, since value callbacks will initialize and run callbacks on their own.  Previously we needed to do this for the test suite to run, but switching from AttributeSignal fixed this in the test suite

### 3. `get_lightpath_state` returns None
- **Problem**: occasionally a device would fail to return a `LightpathState` from `get_lightpath_state`, despite devices working in isolation
- **Fix**: Currently just sleeping after LightController creation and before LightApp creation quenches these errors.  Should probably do something more robust.

### 4. TimoutErrors, pyepics failed to send connection and access rights...
- **Problem**: many PV's fail to connect within the default 1.0s.  And with the huge number of them we can get a lot of errors
- **Fix**: Increase the default timeout settings at the `EpicsSignalBase` level

### 5. `get_lightpath_state()` returns None, and subsequent methods fail
- **Problem**: At startup, many instances of `Unable to determine device state`.  The GUI will open, with many devices in an indeterminate (purple) state.  Eventually callbacks will refresh the GUI, and the beampath will be updated.
- **Fix**: Can sleep at startup to let PV's connect and allow devices to fully initialize.  This is probably bad.   This is also not the most breaking issue, it's more annoying than anything. 

### 6. Beamlines that end in the middle of a branch
- **Problem**: #137 .  Some beamlines end in the middle of a branch, and there was no handling of this before
- **Fix**: allow the config to be a dictionary of {branch_name: `end_z`} mappings, and split paths through that branch on the given `end_z`.  

### 7. Upstream checkbox does nothing
- **Fix**: Change behavior to allow user to select a device to filter devices upstream of.  Now that devices live on branches rather than beamlines, there is no natural way to determine what a-priori a device to consider the first on the beamline.  
image

</details>

Tests and fixtures were also touched up, notably to properly prevent devices from being cached by happi, which led to callbacks being carried over between tests.  Despite the short description this one brought me pain.

## Motivation and Context
I want to actually run the lightpath app.

## How Has This Been Tested?
test suite, and lots of interactive testing.

This PR, for now

### DIFF
--- a/lightpath/__main__.py
+++ b/lightpath/__main__.py
@@ -36,6 +36,10 @@ def main(db, hutches):
     db: str
         Path to happi JSON database
     """
+
+    from ophyd.signal import EpicsSignalBase
+    EpicsSignalBase.set_defaults(timeout=10.0, connection_timeout=10.0)
+
     if db is None:
         client = happi.Client.from_config()
     else:
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -98,7 +98,9 @@ class LightController:
         """
         Load a beamline given the facility graph.  Finds all possible
         paths from facility sources to the endstation's branch.
-        Branches are mapped to endstations in the config
+        Branches are mapped to endstations in the config.
+        Each branch can be optionally mapped to a final z to
+        consider part of the path.
 
         Loads valid beampaths into the LightController.beamlines
         attribute for latedr access.
@@ -133,7 +135,19 @@ class LightController:
             devices = [dat['dev'] for _, dat in subgraph.nodes.data()
                        if dat['dev'] is not None]
             bp = BeamPath(*devices, name=endstation)
-            self.beamlines[endstation].append(bp)
+
+            try:
+                # access the last z for this branch
+                # end_branches is a dictionary
+                last_z = end_branches[path[-1]]
+                if last_z:
+                    # append path with all devices before last z
+                    self.beamlines[endstation].append(bp.split(last_z)[0])
+                else:
+                    self.beamlines[endstation].append(bp)
+            except TypeError:
+                # end_branches is a simple list, no splitting needed
+                self.beamlines[endstation].append(bp)
 
     @staticmethod
     def imped_z(path: BeamPath) -> float:
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -141,18 +141,23 @@ class LightController:
                        if dat['dev'] is not None]
             bp = BeamPath(*devices, name=endstation)
 
-            try:
+            if isinstance(end_branches, dict):
                 # access the last z for this branch
-                # end_branches is a dictionary
+                # if end_branches is Dict[BranchName, end_z], grab last
+                # allowable z with the branch name (last node in path)
                 last_z = end_branches[path[-1]]
                 if last_z:
                     # append path with all devices before last z
                     self.beamlines[endstation].append(bp.split(last_z)[0])
                 else:
                     self.beamlines[endstation].append(bp)
-            except TypeError:
+            elif isinstance(end_branches, list):
                 # end_branches is a simple list, no splitting needed
                 self.beamlines[endstation].append(bp)
+            else:
+                raise TypeError('config is incorrectly formatted '
+                                f'(found mapping from {endstation} to '
+                                f'{type(end_branches)}).')
 
     @staticmethod
     def imped_z(path: BeamPath) -> float:
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -14,6 +14,7 @@ from typing import Any, Dict, List, Set, Tuple
 
 import networkx as nx
 from happi import Client, SearchResult
+from networkx.exception import NodeNotFound
 from ophyd import Device
 
 from .config import beamlines, sources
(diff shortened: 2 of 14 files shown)


tangkong:
Not the longest lines-of-code-wise, but there are enough separate ideas/issues addressed here that I'd like to close the loop on this before moving on to other issues.  The LCLS lightpath now opens and updates for all the beamlines, though it keeps a few non-breaking errors/output spam.

lightpath/__main__.py
```diff
@@ -36,6 +36,10 @@ def main(db, hutches):
     db: str
         Path to happi JSON database
     """
+
+    from ophyd.signal import EpicsSignalBase
+    EpicsSignalBase.set_defaults(timeout=10.0, connection_timeout=10.0)
```
klauer, line 41: This seems... fine. 

What are your thoughts on if we made this configurable as an environment variable-adjustable setting? `LIGHTPATH_EPICS_TIMEOUT` or perhaps even better, if done at the pcdsdevices level, `OPHYD_EPICS_TIMEOUT` or something like that

lightpath/controller.py
```diff
@@ -119,21 +122,42 @@ def load_beamline(self, endstation: str):
...
 
         self.beamlines[endstation] = []
         for path in paths:
             subgraph = self.graph.subgraph(path)
             devices = [dat['dev'] for _, dat in subgraph.nodes.data()
                        if dat['dev'] is not None]
             bp = BeamPath(*devices, name=endstation)
-            self.beamlines[endstation].append(bp)
+
+            if isinstance(end_branches, dict):
```
klauer, line 144: This seems worthy of pulling out into a utility function (`end_branch_to_last_z(end_branches, path[-1])` or something)

lightpath/tests/conftest.py
```diff
@@ -306,7 +290,19 @@ def lcls():
 def lcls_client():
     db = os.path.join(os.path.dirname(__file__), 'path.json')
     print(db)
-    return happi.Client(path=db)
+    client = happi.Client(path=db)
+
+    # monkeypatch to never use cache to get device
+    # cached devices will retain callback information between tests
+    old_get = happi.SearchResult.get
```
klauer, line 297: Would recommend using the `monkeypatch` pytest fixture; it's pretty nice and handles the cleanup for you

`monkeypatch.setattr(happi.SearchResult, "get", new_get)`

(oh, I see you used it elsewhere... Well, the recommendation stands regardless)

lightpath/ui/gui.py
```diff
@@ -263,6 +271,13 @@ def update_path(self, *args,  **kwargs):
                 for widget in row:
                     widget.update_light(_in, _out)
 
+    def _destroy_lightpath_summary_signals(self, *args, **kwargs):
+        """ Update all widgets in rows """
+        # destroy all signals
+        logger.debug('destroying all lightpath_summary signals')
+        for row in self.rows:
+            row[0].device.lightpath_summary.destroy()
```
klauer, line 279: Destroy is pretty heavy-handed as your PR text indicates; could just unsubscribing from subscriptions suffice? Do we need an atef-like [on-exit cleanup](https://github.com/pcdshub/atef/blob/a7c875f7f31c81a00eb174e43113866f05c682ec/atef/util.py#L19-L23)? (And heck, should that be in pcdsutils at this point?)

klauer (approved):
All of this looks good. (If, by chance, you agree with any of my thoughts/suggestions, they probably should come in a follow-up PR.)

lightpath/__main__.py
tangkong: It's a bit annoying since it seems to be only lightpath that runs into these issues at this point.  I'd say if we find any other cases where we want to increase the timeout, it might be worth making this into an environment variable. 

Otherwise I'm on board with pulling this into an environment variable, maybe we could set that in the conda environment?

lightpath/tests/conftest.py
tangkong: I don't know why I didn't do this..

lightpath/ui/gui.py
tangkong: You're right, I think `unsubscribe_all` should work.  `destroy()` sounds angry, but in this case is actually just unsubscribing from all the signals the `SummarySignal` watches.  

The atef cleanup reaches into the control layer and stops it directly, which is more widespread than I think we need here, since we know exactly where the callbacks are. It does seem like something that would be useful in general though, and it'd have been good to know that was an option.

lightpath/controller.py
tangkong: I think I tend to only pull things out into a utility function if I use it more than once, but it would make things more readable.

klauer: Yeah, I like the readability of utility functions - I've tended to put more in nowadays than I used to, and I think my code style has improved because of it.

ZLLentz (approved):
I have nothing specific to add- Ken covered anything I saw and more. :+1:

Issue #143, opened 2022-07-28
ENH: Make device loading more lazy


tangkong:
## Expected Behavior
It'd be nice if we don't have to load all devices in the facility when initializing lightpath. There should be a way to only create the devices we care about, even if we're building the whole facility graph to find paths.

I had this in mind when creating the lightpath interface, but didn't fully leverage it.

## Current Behavior
Currently on the creation of the facility graph, we load all the devices.  This makes it simpler to put these devices into their beampaths, but also incurs a lot of startup cost. 

## Possible Solution
- Load the facility graph and stash the SearchResult instead of the device
- Make `LightController.beamlines` hold paths (lists of nodes) instead of `BeamPath`
- Create an access-er method that fills a `BeamPath` if it does not already exist, and returns a BeamPath if it has been filled.
- Need to check when the gui subscribes to beampaths.  Make sure `BeamPaths` are loaded before subscriptions occur.

## Context
General startup costs must be squashed at all costs

## Your Environment
pcds-5.4.1, with the updated lightpath-device-config

tangkong:
closed by #145

Issue #139, opened 2022-07-13
Make device handling more robust


tangkong:
## Expected Behavior
1. Figure out how to best classify devices as "upstream" and apply this to `lightpath.gui.LightApp.filter`

2. Properly handle devices that fail to instantiate. (`lightpath.controller.LightController.make_graph`)

## Current Behavior
Currently the gui will just take the whole active path, need to figure out how to best handle upstream and downstream devices

## Possible Solution
1: added logic. might be sufficient on the gui side of things?
2: Possibly make a dummy device that spoofs the missing device?  Mimic the container's input / output devices?

## Context
Left behind from #133 

## Your Environment
pcds-5.4.1 + #133

tangkong:
Part 1 addressed in #142

tangkong:
Closed by the aforementioned PRs 🎉

PR #145, opened 2022-08-01
ENH: Make device loading lazy, make mock devices available

tangkong:
## Description
- Make the facility graph use only database information
- Defer device instantiation until needed (when making BeamPath objects, looking for beam destinations etc)
- Attempt to make a mock device if device instantiation fails initially
  - Move simulated devices from test suite into separate module for package-wide use

## Motivation and Context
#143 
#139 part 2.

## How Has This Been Tested?
Test Suite runs, extra smoke test for mocked devices added. 

This PR

### DIFF
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -18,7 +18,7 @@ from networkx.exception import NodeNotFound
 from ophyd import Device
 
 from .config import beamlines, sources
-from .errors import PathError
+from .errors import DeviceLoadingError, PathError
 from .path import BeamPath
 
 logger = logging.getLogger(__name__)
@@ -138,15 +138,20 @@ class LightController:
 
         self.beamlines[endstation] = paths
 
-    def get_paths(self, endstation):
+    def get_paths(self, endstation: str) -> List[BeamPath]:
         """
         Returns the BeamPaths for a specified endstation.
         Create and fill the BeamPaths if they have not been already
 
         Parameters
         ----------
-        endstation : _type_
-            _description_
+        endstation : str
+            name of endstation to return paths for
+
+        Returns
+        -------
+        List[BeamPath]
+            a list of BeamPath's to the requested endstation
         """
         paths = self.beamlines[endstation]
 
@@ -459,14 +464,6 @@ class LightController:
         # label nodes with device name, store ophyd device
         nodes = []
         for res in result_list:
-            # try:
-            #     dev = res.get()
-            # except Exception:
-            #     # TODO: be better about specific exceptions
-            #     logger.debug(
-            #         f'Failed to initialize device: {res["name"]}'
-            #     )
-            #     continue
             nodes.append((res.metadata['name'],
                          {'res': res, 'dev': None}))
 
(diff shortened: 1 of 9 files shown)


lightpath/controller.py
```diff
@@ -432,15 +465,8 @@ def make_graph(
...
-                dev = res.get()
-            except Exception:
-                # TODO: be better about specific exceptions
-                logger.debug(
-                    f'Failed to initialize device: {res["name"]}'
-                )
-                continue
-            nodes.append((res.metadata['name'], {'dev': dev}))
+            nodes.append((res.metadata['name'],
+                         {'res': res, 'dev': None}))
```
ZLLentz, line 469: So this is the core of the PR here? Store the happi result with each node and use the happi result whenever you want to check the metadata instead of needing to build all the devices. I think this is a good choice and it looks like you propagated it everywhere.

lightpath/controller.py
```diff
@@ -432,15 +465,8 @@ def make_graph(
         # label nodes with device name, store ophyd device
```
ZLLentz, line 465: Outdated comment, maybe

lightpath/controller.py
```diff
@@ -527,8 +555,59 @@ def get_device(self, device: NodeName) -> Device:
         -------
         Device
             requested device
+
+        Raises
+        ------
+        DeviceLoadingError
+            if the requested device is in the facility but cannot be loaded
```
ZLLentz, line 562: Is this docstring true? It looks like the new behavior is to slot in a mock device on failure

lightpath/controller.py
```diff
@@ -527,8 +555,59 @@ def get_device(self, device: NodeName) -> Device:
...
+        elif dev_data['res'] is not None:
+            # not instantiated yet, create and fill
+            try:
+                dev = dev_data['res'].get()
+                self.graph.nodes[device_name]['dev'] = dev
+                return dev
+            except Exception:
+                logger.error(f'Device {device_name} failed to load, '
+                             'attempting to make a mock device')
+                dev = make_mock_device(dev_data['res'])
```
ZLLentz, line 581: Does this mean that unloadable devices will have a stand-in node? If so, what do these stand-in nodes look like to the user? Are they blocking, for example?

ZLLentz (approved):
I think this is pretty solid, added questions to verify my understanding and poke at the docstrings a bit.

lightpath/controller.py
tangkong: Yep, you got it.  As usual a bunch of other things crept in before I told myself to stop

lightpath/controller.py
tangkong: Ah you're right.  I took this out after adding the mock devices.  I'll fix that.

lightpath/controller.py
tangkong: They will.  Currently they're set to be removed, since I figured if we have a stand-in node it shouldn't be disruptive.  If we wanted to assume the beam is blocked we could just go back to throwing an error if we can't create the device.  (both would force users to go and fix the device class/happi information/whatever issue)

Though it's probably worth making these nodes very obvious to the user so they know something is wrong.  I'll throw a screenshot up, but currently I just rename the device and have it operate as a simulated valve/crystal.

I'm open to suggestions on this though, the more I think about it the more I move toward the fence.

ZLLentz: I think your philosophy of being minimally disruptive makes sense. It's possible the main things we can do are startup warnings and assigning some unique color to it... but my initial thought about faking a blocked state I think was misguided. The other thing we might be able to do is pick a silly icon for them so they stand out in the GUI.

tangkong: I need to go through and make the coloring consistent overall as well (inserted mirrors with 1.0 transmission shouldn't be red, etc), so I'll touch up the stand-in gui elements there.

klauer: I like the idea of an icon that would stand out somehow

lightpath/controller.py
```diff
@@ -484,28 +511,30 @@ def make_graph(
                 last_on_branch = i
 
             try:
-                next_dev = nodes[i+1][1]['dev']
+                next_dev = nodes[i+1][1]['res']
```
klauer, line 514: Seems like with the `res`/`dev` stuff it might be worthwhile having a single keyed set of information:
```
@dataclass
class NodeMetadata:
    res: SearchResult
    dev: Optional[ophyd.Device]


node[...]["md"] = NodeMetadata(...)
```

klauer (approved):
Looking good!

lightpath/controller.py
tangkong: I like this

PR #148, opened 2022-08-03
FIX: clear subscriptions properly

tangkong:
## Description
Admittedly a bunch of bug fixes that should have been caught when we changed the subscription schema
- Properly unsubscribes callbacks from the `lightpath_summary` signal, instead of the `BeamPath` object.  
- destroy the lightpath_summary signals in all the widgets, not just the first half
- Creates `LightApp.clear_device_subs()` to unsub `BeamPath._device_moved()` all devices in a BeamPath.

## Motivation and Context
This fixes a bug but I went down too many rabbit holes and forgot to document fully with screenshots.  It definitely helps with subscriptions cleanly though.

## How Has This Been Tested?
Interactively

This PR

### DIFF
--- a/lightpath/path.py
+++ b/lightpath/path.py
@@ -556,6 +556,17 @@ class BeamPath(OphydObject):
             self._has_subscribed = True
         super().subscribe(cb, event_type=event_type, run=run)
 
+    def clear_device_subs(self) -> None:
+        """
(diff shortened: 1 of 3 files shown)


klauer (approved):

ZLLentz (approved):
Looks solid to me, good find. It's hard to fully understand the repercussions but it looks sensible

PR #150, opened 2022-08-16
GUI: More expressive widget icon colors, and more along the way

### LINKED ISSUE #144: GUI: Fix device type filters
## Expected Behavior
Device type filter checkboxes should properly hide/show devices

## Current Behavior
Check boxes currently don't work

## Possible Solution
The checkboxes filter on happi container, but should probably filter on device_class.
It might be worth populating these more dynamically?  Currently we only provide the option to filter a fixed set of device types
Some devices are subclasses of others, so maybe we group those together somehow?

## Your Environment
pcds-5.4.1

tangkong:
## Description
We're finally getting into the issues from the jira tickets!  Kinda.

- Gives the GUI a title that is not "form"
- Removes sorting of endstations to allow lazy loading of BeamPaths
- Fixes device filter area to match with `isinstance`, and no longer use factory functions from pcdsdevices. 
  - closes #144 
- Track cumulative transmission along BeamPath, and mark devices as blocking if they force total transmission below a threshold
- Color the widget based on both the device state and path state.   
  - A variety of jira tickets, including [jira1](https://jira.slac.stanford.edu/browse/LCLSECSD-143), [jira2](https://jira.slac.stanford.edu/browse/LCLSPC-135), [jira3](https://jira.slac.stanford.edu/browse/LCLSECSD-748)

Also in a more minor sense
- Fixes test suite fa4 deprecation warnings
- updates some docstrings and type hints
- updates tests
  - tests for attenuation tracking and widget color updates
  - un-hard-codes `path.show_devices()` test, using regex and pain
- makes the lcls simulated facility more accessible

## Motivation and Context
closes #144 , and a few jira tickets as detailed above

<details>
  <summary>Thoughts and justifications of my decisions, to make myself feel better</summary>
  
- Accumulated transmission logic: Inserted devices can now have non-0 transmission, but should removed devices have the ability to block?
  - Currently the logic already assumes that removed devices cannot be blocking, unless there is branch mismatch
  - While it's not absurd to imagine a device that blocks when removed, I chose to punt the issue
- Now that the `LightRow` widget relies on the path and the device, are we going to have a race condition?
  - Two `LightRow` widgets and one `BeamPath` are subscribed to changes in `device.lightpath_summary`.
  - When `lightpath_summary` changes, is it possible that the `LightRow`'s update before the `BeamPath` finishes, giving us bad state information?
  - No, callbacks are processed in insertion-order, and the `BeamPath` callbacks are subscribed before the `LightRow` widgets
  - I should really draw out the subscription structure in a nice way so people don't have to go through the maze repeatedly

</details>

## How Has This Been Tested?
Interactive testing, and tests have been added to the suite.

This PR, some docstrings, etc

## Some example screenshots
From sim mode, sim_ipimb's have an inserted transmission of 0.5, so inserting 4 drops the beam below the minimum of 0.1.  Inserted devices are now blue, and device types have been updated
image

An example from XCS
image

### DIFF
--- a/lightpath/path.py
+++ b/lightpath/path.py
@@ -236,12 +236,11 @@ class BeamPath(OphydObject):
                 block.append(prev_device)
             # check inserted
             elif curr_state is DeviceState.Inserted:
-                current_transmission *= curr_status.transmission
-                # Blocking if single device has low enough transmssion
-                if curr_status.transmission < self.minimum_transmission:
-                    block.append(device)
+                if curr_status.transmission > 1:
+                    logger.error(f'{device.name} reports transmission > 1')
+                current_transmission *= min(curr_status.transmission, 1)
                 # Any device seeing current transmission below min would block
-                elif current_transmission < self.minimum_transmission:
+                if current_transmission < self.minimum_transmission:
                     block.append(device)
             # Do not add device to blocking list, it is removed
             elif curr_state is DeviceState.Removed:
--- a/lightpath/path.py
+++ b/lightpath/path.py
@@ -214,12 +214,14 @@ class BeamPath(OphydObject):
         Returns
         -------
         List[Device]
-            list of blockoing devices
+            list of blocking devices
         """
         # Cache important prior devices
         prev_device = None
         prev_status = None
         block = list()
+        current_transmission = 1
+
         for device in self.path:
             curr_state, curr_status = find_device_state(device)
 
@@ -234,13 +236,17 @@ class BeamPath(OphydObject):
                 block.append(prev_device)
             # check inserted
             elif curr_state is DeviceState.Inserted:
-                # Ignore devices with low enough transmssion
+                current_transmission *= curr_status.transmission
+                # Blocking if single device has low enough transmssion
                 if curr_status.transmission < self.minimum_transmission:
                     block.append(device)
-            # Find unknown and inconsistent devices
+                # Any device seeing current transmission below min would block
+                elif current_transmission < self.minimum_transmission:
+                    block.append(device)
+            # Do not add device to blocking list, it is removed
             elif curr_state is DeviceState.Removed:
                 pass
-            # Do not add device to blocking list, it is removed
+            # Unknown and inconsistent devices block
             else:
                 block.append(device)
 
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -217,12 +217,6 @@ class LightApp(Display):
                 self.rows.clear()
                 self.device_combo.clear()
                 self.upstream_device_combo.clear()
-            # # Hide nothing when switching beamlines
-            # boxes = self.device_types.children()
-            # boxes.extend([self.remove_check])
-            # for box in boxes:
-            #     if isinstance(box, QCheckBox):
-            #         box.setChecked(True)
 
             # Grab all the light rows (self.path set here)
             rows = [self.load_device_row(d)
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -21,6 +21,7 @@ from typhos import TyphosDeviceDisplay
 
 from lightpath.path import DeviceState
 
+from ..mock_devices import IPIMB, Crystal, Stopper
 from .widgets import LightRow
 
 logger = logging.getLogger(__name__)
@@ -52,7 +53,9 @@ class LightApp(Display):
                    'KBOMirror': KBOMirror, 'OffsetMirror': XOffsetMirror,
                    'PIM': dtypes.PIM, 'PPSStopper': PPSStopper,
                    'PulsePicker': dtypes.PulsePicker, 'Slit': dtypes.Slits,
-                   'Stopper': dtypes.Stopper, 'XFLS': dtypes.XFLS}
+                   'Stopper': dtypes.Stopper, 'XFLS': dtypes.XFLS,
+                   'SIM_IPIMB': IPIMB, 'SIM_Stopper': Stopper,
+                   'SIM_Crystal': Crystal}
 
     def __init__(self, controller, beamline=None,
                  parent=None, dark=True):
@@ -92,6 +95,8 @@ class LightApp(Display):
         self.detail_hide.clicked.connect(self.hide_detailed)
         # Store LightRow objects to manage subscriptions
         self.rows = list()
+        # store device type filter widgets
+        self._device_checkboxes = list()
         # Select the beamline to begin with
         beamline = beamline or self.destinations()[0]
         try:
@@ -101,18 +106,6 @@ class LightApp(Display):
             idx = 0
         # Move the ComboBox
         self.destination_combo.setCurrentIndex(idx)
-        # Add all of our device type options
-        max_columns = 3
-        for i, row in enumerate(np.array_split(list(self.shown_types.items()),
-                                               max_columns)):
-            for j, device_type in enumerate(row):
-                # Add box to layout
-                box = QCheckBox(device_type[0])
-                box.setChecked(True)
-                self.device_types.layout().addWidget(box, j, i)
-                # Hook up box to hide function
-                self.device_buttons[box] = device_type[1]
-                box.toggled.connect(self.filter)
         # Setup the UI
         self.change_path_display()
         self.resizeSlider()
@@ -179,6 +172,40 @@ class LightApp(Display):
         return [dtype for button, dtype in self.device_buttons.items()
                 if not button.isChecked()]
 
+    def update_device_types(self):
+        """
+        Update the device type filter check boxes bases on devices
+        currently in the path
+        """
+        with self._lock:
+            # find valid device types. any() short circuits, yay generators!
+            valid_types = {}
+            for type_name, dtype in self.shown_types.items():
+                if any(isinstance(dev, dtype) for dev in self.path.path):
+                    valid_types[type_name] = dtype
+
+            # clear old checkboxes
+            for child in self._device_checkboxes:
+                self.device_types.layout().removeWidget(child)
+                self.device_buttons.pop(child)
+                child.deleteLater()
+
+            self._device_checkboxes.clear()
+
+            # populate grid with available device types
+            max_columns = 3
+            for i, row in enumerate(np.array_split(list(valid_types.items()),
+                                                   max_columns)):
+                for j, device_type in enumerate(row):
+                    # Add box to layout
+                    box = QCheckBox(device_type[0])
+                    box.setChecked(True)
+                    self.device_types.layout().addWidget(box, j, i)
+                    # Hook up box to hide function
+                    self.device_buttons[box] = device_type[1]
+                    box.toggled.connect(self.filter)
+                    self._device_checkboxes.append(box)
+
     @pyqtSlot()
     @pyqtSlot(bool)
     def change_path_display(self, value=None):
@@ -213,7 +240,7 @@ class LightApp(Display):
                 if isinstance(box, QCheckBox):
                     box.setChecked(True)
 
-            # Grab all the light rows
+            # Grab all the light rows (self.path set here)
             rows = [self.load_device_row(d)
                     for d in self.select_devices(self.selected_beamline())]
             # Add all the widgets to the display
@@ -240,6 +267,8 @@ class LightApp(Display):
                 widget.update_state()
         # Update the state of the path
         self.update_path()
+        # Update device type checkboxes
+        self.update_device_types()
 
     def ui_filename(self):
         """
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -11,6 +11,7 @@ import pcdsdevices.device_types as dtypes
 import typhos
 from pcdsdevices.attenuator import AttBase
 from pcdsdevices.ipm import IPMMotion
+from pcdsdevices.mirror import KBOMirror, XOffsetMirror
 from pcdsdevices.valve import PPSStopper
 from pydm import Display
 from qtpy.QtCore import Qt
@@ -46,10 +47,12 @@ class LightApp(Display):
 
     parent : optional
     """
-    shown_types = [AttBase, dtypes.GateValve, IPMMotion,
-                   dtypes.LODCM, dtypes.OffsetMirror, dtypes.PIM, PPSStopper,
-                   dtypes.PulsePicker, dtypes.Slits, dtypes.Stopper,
-                   dtypes.XFLS]
+    shown_types = {'Attenuator': AttBase, 'Gate Valve': dtypes.GateValve,
+                   'IPM': IPMMotion, 'LODCM': dtypes.LODCM,
+                   'KBOMirror': KBOMirror, 'OffsetMirror': XOffsetMirror,
+                   'PIM': dtypes.PIM, 'PPSStopper': PPSStopper,
+                   'PulsePicker': dtypes.PulsePicker, 'Slit': dtypes.Slits,
+                   'Stopper': dtypes.Stopper, 'XFLS': dtypes.XFLS}
 
     def __init__(self, controller, beamline=None,
                  parent=None, dark=True):
@@ -100,15 +103,15 @@ class LightApp(Display):
         self.destination_combo.setCurrentIndex(idx)
         # Add all of our device type options
         max_columns = 3
-        for i, row in enumerate(np.array_split(self.shown_types,
+        for i, row in enumerate(np.array_split(list(self.shown_types.items()),
                                                max_columns)):
             for j, device_type in enumerate(row):
                 # Add box to layout
-                box = QCheckBox(device_type.__name__)
+                box = QCheckBox(device_type[0])
                 box.setChecked(True)
                 self.device_types.layout().addWidget(box, j, i)
                 # Hook up box to hide function
-                self.device_buttons[box] = device_type
+                self.device_buttons[box] = device_type[1]
                 box.toggled.connect(self.filter)
         # Setup the UI
         self.change_path_display()
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -66,6 +66,7 @@ class LightApp(Display):
         self.detail_screen = None
         self.device_buttons = dict()
         self._lock = threading.Lock()
+        self._prev_block = None
         # Create empty layout
         self.lightLayout = QHBoxLayout()
         self.lightLayout.setSpacing(1)
@@ -310,8 +311,10 @@ class LightApp(Display):
                 for widget in row:
                     widget.update_light(_in, _out)
                     # Reconsider blocking device state
-                    if device is block:
+                    if (device is self._prev_block or device is block):
                         widget.update_state()
+
+            self._prev_block = block
 
     def _destroy_lightpath_summary_signals(self, *args, **kwargs):
         """ Update all widgets in rows """
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -21,7 +21,7 @@ from typhos import TyphosDeviceDisplay
 
 from lightpath.path import DeviceState
 
-from ..mock_devices import IPIMB, Crystal, Stopper
+from ..mock_devices import IPIMB, Crystal, Stopper, Valve
 from .widgets import LightRow
 
 logger = logging.getLogger(__name__)
@@ -55,7 +55,7 @@ class LightApp(Display):
                    'PulsePicker': dtypes.PulsePicker, 'Slit': dtypes.Slits,
                    'Stopper': dtypes.Stopper, 'XFLS': dtypes.XFLS,
                    'SIM_IPIMB': IPIMB, 'SIM_Stopper': Stopper,
-                   'SIM_Crystal': Crystal}
+                   'SIM_Crystal': Crystal, 'SIM_Valve': Valve}
 
     def __init__(self, controller, beamline=None,
                  parent=None, dark=True):
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -124,10 +124,9 @@ class LightApp(Display):
 
     def destinations(self):
         """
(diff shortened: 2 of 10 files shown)


tangkong:
This is probably ready for some eyes. The exact names and colors I used will surely be changed in the future, but hopefully this PR provides us an opportunity to customize such things.

I had thoughts of adding the option to switch between various color maps, but that's maybe for a different pr.

ZLLentz:
I'll give this a look! Any special instructions for running it myself? Do I need to set up a pcdsdevices dev too?

lightpath/path.py
```diff
@@ -214,12 +214,14 @@ def blocking_devices(self) -> List[Device]:
         Returns
         -------
         List[Device]
-            list of blockoing devices
```
ZLLentz, line 217: This old typo gave me a giggle

lightpath/path.py
```diff
@@ -234,13 +236,17 @@ def blocking_devices(self) -> List[Device]:
                 block.append(prev_device)
             # check inserted
             elif curr_state is DeviceState.Inserted:
-                # Ignore devices with low enough transmssion
+                current_transmission *= curr_status.transmission
```
ZLLentz, line 239: Do we need a check here for `curr_status.transmission` <= 1? Maybe an exception or a min(transmission, 1) clamp?

lightpath/path.py
```diff
@@ -234,13 +236,17 @@ def blocking_devices(self) -> List[Device]:
...
+                # Blocking if single device has low enough transmssion
                 if curr_status.transmission < self.minimum_transmission:
                     block.append(device)
-            # Find unknown and inconsistent devices
+                # Any device seeing current transmission below min would block
+                elif current_transmission < self.minimum_transmission:
+                    block.append(device)
```
ZLLentz, line 245: Does this new total transmission condition make the old single device condition redundant?

Note that this is a change in behavior from previous- I think it's more useful this way

lightpath/ui/gui.py
```diff
@@ -173,6 +172,40 @@ def hidden_devices(self):
         return [dtype for button, dtype in self.device_buttons.items()
                 if not button.isChecked()]
 
+    def update_device_types(self):
+        """
+        Update the device type filter check boxes bases on devices
+        currently in the path
```
ZLLentz, line 178: This is a good idea, it was always weird before having useless filters loaded

lightpath/ui/gui.py
```diff
@@ -44,10 +48,14 @@ class LightApp(Display):
...
-                   dtypes.PulsePicker, dtypes.Slits, dtypes.Stopper,
-                   dtypes.XFLS]
+    shown_types = {'Attenuator': AttBase, 'Gate Valve': dtypes.GateValve,
+                   'IPM': IPMMotion, 'LODCM': dtypes.LODCM,
+                   'KBOMirror': KBOMirror, 'OffsetMirror': XOffsetMirror,
+                   'PIM': dtypes.PIM, 'PPSStopper': PPSStopper,
+                   'PulsePicker': dtypes.PulsePicker, 'Slit': dtypes.Slits,
+                   'Stopper': dtypes.Stopper, 'XFLS': dtypes.XFLS,
+                   'SIM_IPIMB': IPIMB, 'SIM_Stopper': Stopper,
+                   'SIM_Crystal': Crystal}
```
ZLLentz, line 58: Do we have to consider the placeholder items here for the devices that didn't load properly? Or is that baked in already?

lightpath/ui/gui.py
```diff
@@ -274,6 +309,9 @@ def update_path(self, *args,  **kwargs):
                 # Update widget display
                 for widget in row:
                     widget.update_light(_in, _out)
+                    # Reconsider blocking device state
+                    if device is block:
+                        widget.update_state()
```
ZLLentz, line 314: Is this an optimization of some sort? Or is it a bugfix?

lightpath/ui/widgets.py
```diff
@@ -18,12 +18,15 @@
...
 # Define the state colors that correspond to DeviceState
-state_colors = [QColor(124, 252, 0),  # Removed
-                QColor(255, 0, 0),  # Inserted
-                QColor(255, 215, 0),  # Unknown
-                QColor(255, 215, 0),  # Disconnected
-                QColor(255, 0, 255),  # Disconnected
-                QColor(255, 0, 255)]  # Error
+state_colors = {
+    'removed': QColor(124, 252, 0),  # Removed (green)
+    'half_removed': QColor(0, 176, 255),  # half-removed (light blue)
```
ZLLentz, line 23: Is this the "inserted but not blocking" state?

lightpath/ui/widgets.py
```diff
@@ -177,21 +180,52 @@ def add_signal(self, signal):
         self.command_layout.insertWidget(count - 1, label, 0, Qt.AlignHCenter)
         self.command_layout.insertWidget(count, widget, 0, Qt.AlignHCenter)
 
+    def get_state_color(self) -> QColor:
+        """
+        Determine the icon color given the device state and path status
+        If device is an impediment: StateColor.Blocking
+        If device is removed: StateColor.Removed
+        If device is inserted, not blocking (mirrors): StateColor.HalfRemoved
+        If device is unknown / errored: StateColor.Error
```
ZLLentz, line 189: Docstring here is left over from a previous iteration of the code where the state colors were an enum, I think?

ZLLentz (commented):
This looks great to me
Added questions and comments to make sure every case is considered

lightpath/ui/widgets.py
tangkong: Yep, I'm sure there's a better name but that's what I went with for now.

lightpath/ui/gui.py
tangkong: Ah this was actually left over from when I colored downstream half-removed devices as half-removed, not blocking.  However this does remind me of a bug I'll fix shortly:

If you insert a bunch of attenuating devices, the last device gets colored red as expected.  But if you remove an upstream device and the final blocking device changes, the previous blocking device needs to change.

tangkong:
> I'll give this a look! Any special instructions for running it myself? Do I need to set up a pcdsdevices dev too?

You'll need updates from 3 branches:
* The most recent `lightpath` branch/PR
* The LightpathMixin `pcdsdevices` PR: https://github.com/pcdshub/pcdsdevices/pull/1028
* The updated `device_config` implementing LightpathMixin on the happi database: https://github.com/pcdshub/device_config/pull/31

Once those three are in your path, you should be able to redirect `HAPPI_CFG` to the new device_config db and load lightpath.  Loading `lightpath --sim` should work independent of that

lightpath/ui/widgets.py
tangkong: Ah right, they're dictionary access now.

lightpath/ui/gui.py
tangkong: Ah, I didn't include those.  I can do that.

lightpath/path.py
tangkong: I think this is also leftover from when I colored downstream half-removed devices as half-removed, not blocking.  It is redundant now.

lightpath/ui/widgets.py
ZLLentz: I also have no better suggestion despite it seeming like one should exist

ZLLentz:
It looks to me like I also need to include `lightpath` as a pip -e install to make the container work right and I need to include the master branch of `happi` too

ZLLentz:
I get strange behavior when I run this: lots of error spam and then included device types is surprisingly bare. I suspect maybe I haven't set up my environment correctly but I can't tell. The new colors show up and look good.
image

tangkong:
Yeah, I'm working out the fix to the error spam now.  It's been bothering me for quite a while, and I think I made a breakthrough (stay tuned).

The device types list is bare because we're looking for specific pre-determined device types, and pcdsdevices has become quite... varied in their inheritance.  (e.g., at2l0 doesn't inherit from `AttBase`).  Something I wanted to do was populate that table with the existing devices, but couldn't immediately think of a clean way to decide what the "parent" of a device should be, or when to stop grouping devices together (a lot of devices might be reasonably grouped as "positioners", but not all of them?).  We could also let the user configure this, but we'd have to build up the machinery for that.  I remember thinking about this and eventually also punting it. 

If all this looks good I can make an issue

lightpath/ui/gui.py
tangkong: Actually, I didn't forget it before (I just forgot to write down why I made that decision).  Because all the sim devices inherit from `Valve`, unchecking it removes all the sim devices.  While in sim-mode you can just put the devices back, in normal mode we mock devices with either `Valve` or `Crystal`, the latter of the two inheriting from `Valve`.  🤔

ZLLentz:
> The device types list is bare because we're looking for specific pre-determined device types, and pcdsdevices has become quite... varied in their inheritance.

Maybe this should become its own issue if there is enough unrelated stuff in this PR. I'd expect that I'd be able to make all of the devices disappear by clicking all the filter buttons, but that doesn't seem to be possible (it wasn't possible in the old lightpath either, to be fair)

`lightpath --sim` doesn't load properly for me which makes me think I've made an error in my dev setup here...

ZLLentz:
I don't think this closes #144 though. The buttons still don't work fully, there are still valves that remain when clicking the valves button for example.

lightpath/ui/gui.py
```diff
@@ -44,10 +48,14 @@ class LightApp(Display):
 
     parent : optional
     """
-    shown_types = [dtypes.Attenuator, dtypes.GateValve, dtypes.IPM,
-                   dtypes.LODCM, dtypes.OffsetMirror, dtypes.PIM, PPSStopper,
-                   dtypes.PulsePicker, dtypes.Slits, dtypes.Stopper,
-                   dtypes.XFLS]
+    shown_types = {'Attenuator': AttBase, 'Gate Valve': dtypes.GateValve,
```
ZLLentz, line 51: `isinstance` can check against tuples of types, so we could include tuples of types in the dict here to catch cases where multiple unrelated type trees represent the same "kind" of object.

Alternatively, maybe it's time to rework pcdsdevices to include bare `AttenuatorType` base classes with no code in them.

lightpath/ui/gui.py
ZLLentz: But surely there should be a way in non-sim mode to hide the mock devices? Either with their own button or with the original button for that class?

tangkong:
> I don't think this closes #144 though. The buttons still don't work fully, there are still valves that remain when clicking the valves button for example.

Which valves remain?  the SIM_Valve only refers to the mock device, and I think not all valves are GateValves?

ZLLentz:
Pretty much all valves in the lightpath are gate valves

ZLLentz:
From this screenshot:

tv1l0_vgc01 should be grouped with the gate valves (class `VGC` = Valve of Gate type Controlled)
im1l0 should be grouped with the pims (class `XPIM` = XTES PIM)
sl1l0 should be grouped with the slits (class `PowerSlits`)

image

I think we should sit down and try to put all the remaining beamline objects into one of the existing boxes

lightpath/ui/gui.py
tangkong: Agreed, I just need to change the mock classes so Crystal isn't also Valve

ZLLentz:
Lingering combobox error spam can be sourced to https://github.com/pcdshub/pcdsdevices/pull/1046

lightpath/ui/gui.py
klauer: I'd be OK with that or just type-hint based ones: `AttenuatorType = Union[AttBase, ...]`

klauer: This could be dynamically generated based on the device instances that are currently shown, though it wouldn't be as nice as a manually curated list.

lightpath/ui/gui.py
```diff
@@ -96,39 +107,28 @@ def __init__(self, controller, beamline=None,
...
-                self.device_buttons[box] = device_type
-                box.toggled.connect(self.filter)
         # Setup the UI
         self.change_path_display()
         self.resizeSlider()
         # Change the stylesheet
         if dark:
             typhos.use_stylesheet(dark=True)
 
+        self.setWindowTitle('Lightpath')
```
klauer, line 117: Title change, yes! 😁 

```suggestion
        self.setWindowTitle(f'Lightpath - {beamline}' if beamline else "Lightpath")
```

lightpath/ui/gui.py
```diff
@@ -173,6 +173,40 @@ def hidden_devices(self):
...
+            valid_types = {}
+            for type_name, dtype in self.shown_types.items():
+                if any(isinstance(dev, dtype) for dev in self.path.path):
+                    valid_types[type_name] = dtype
+
+            # clear old checkboxes
+            for child in self._device_checkboxes:
+                self.device_types.layout().removeWidget(child)
+                self.device_buttons.pop(child)
+                child.deleteLater()
```
klauer, line 192: I know there was talk of a double free-related segfault. I admittedly didn't pay enough attention to know if it could be related, but I see `deleteLater` here...

You could avoid deleting these by just creating all of the checkboxes and hiding the ones that aren't useful to avoid a create/remove/repeat

lightpath/ui/widgets.py
```diff
@@ -177,21 +180,52 @@ def add_signal(self, signal):
...
+        -------
+        QColor
+            the color to apply to the icon
+        """
+        device_state, _ = find_device_state(self.device)
+        blocking_devices = self.path.blocking_devices
+
+        if device_state is DeviceState.Disconnected:
+            return state_colors['disconnected']
+        elif device_state is DeviceState.Error:
```
klauer, line 206: Minor style suggestion - big blocks containing multiple `elif` aren't really necessary when there's `return` in each of the clauses:
```suggestion
        if device_state is DeviceState.Error:
```
Then `else` even becomes unnecessary

lightpath/ui/gui.py
tangkong: Zach and I settled on taking the module name as the "device type", which simplifies the work here and helps incorporate future changes more easily.  

Under this scheme, we'd group all devices from `pcdsdevices.attenuators` under the same checkbox, everything under `pcdsdevices.pim` together, etc

(I'll commit when I'm able, but do you have any objections to this approach?)

klauer: Sounds good to me 👍

lightpath/ui/gui.py
tangkong: I could make it so we never actually delete these checkboxes, rather we just store them and pick the ones we need.  

Is it generally bad to `deleteLater`?  I figured it'd help free up memory and guard against the strange off chance that we accumulate too many device types, but I could try to avoid it in general.

klauer: If I hadn't vaguely recalled about the double free, I wouldn't have even commented. Did you by chance get to the bottom of that issue?

Normally `deleteLater` is just fine and isn't a reason on its own to suggest not using it.  There are scenarios where the overhead of creating/deleting/creating can be worse than just creating everything and hiding what you don't need, though it on a case-by-case basis where either method would be faster/more efficient/preferable.

tangkong: I poked around a bunch trying to reproduce the error, but to no avail.  

Just wanted to make sure I wasn't missing anything canonical w.r.t. pyqt gui development, and I'll certainly keep an eye out for the double free errors

tangkong:
I think I've addressed the concerns here.  I spent this morning poking around the gui trying to illicit a "double free" error, but to no avail.  Instead I:

- found one small bug that was fixed by removing code that modified checkboxes outside of the new `update_device_types` method
- ran into a segmentation fault I could also not replicate

ZLLentz:
Do you want a re-review and/or a re-beta tester?

tangkong:
> Do you want a re-review and/or a re-beta tester?

A re-review would be nice, the main thing that's changed is the device types filtering.  Someone else poking around with it would also be appreciated, maybe the sporadic errors/crashes will show up for others...

Once I merge this I have a documentation rewrite in the pipeline, after which maybe we can tag? 👀

ZLLentz:
Looks good interactively! I'll poke through the source now.
There are a few devices that filter on an unintuitive button but I think that means they need happi/pcdsdevices updates rather than anything special here.

ZLLentz:
I hit a segmentation fault but it wasn't the double free. I'll run again and turn core dumping on while I browse the code.

lightpath/ui/widgets.py
```diff
@@ -177,21 +180,52 @@ def add_signal(self, signal):
...
+        while providing useful information about each device's state.
+
+        Could take a color map in the future?  Colorblind support?
+
+        Returns
+        -------
+        QColor
+            the color to apply to the icon
+        """
+        device_state, _ = find_device_state(self.device)
```
ZLLentz, line 201: I'm not 100% sure about this, but I think this bit might make the lightpath first load go painfully slow when devices are disconnected?

Something is pathing through https://github.com/pcdshub/lightpath/blob/05a9394e09acf256649738f6233d54a7ba2b1b4d/lightpath/path.py#L83 at startup for all devices

Might be good to look at this in a follow-up

ZLLentz: What I'm trying to say is: all disconnected devices are timing out at one of these function calls in series before any device is shown in lightpath

ZLLentz (approved):

lightpath/ui/widgets.py
ZLLentz: It might worse than this actually, I can't seem to switch to a path that has a disconnected device
I just see this forever and my ui freezes
```
[2022-08-24 16:36:19] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:21] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:29] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:31] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:39] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:41] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:49] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:51] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:36:59] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:01] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:09] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:11] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:19] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:21] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:29] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:31] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:39] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:41] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:49] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:51] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:37:59] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:38:01] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:38:09] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:38:11] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
[2022-08-24 16:38:19] - WARNING -  Unable to connect to BeckhoffSlits(SL2K2:SCATTER, name=sl2k2)
```

tangkong: This makes sense.  I think the real root of the issue is the fact that I increased the default EpicsSignal timeout to 10s, otherwise we'd get a slew of timeout errors at startup.  

Each `LightRow` widget updates its state once after they are created, which calls `find_device_state`.  

Something I considered doing was simply making timeout errors less severe, and assuming that devices would connect eventually.  That seemed dangerous though.

ZLLentz: > Each LightRow widget updates its state once after they are created, which calls find_device_state.

Somehow there must be at least one other place this is done that blocks qt from doing anything forever

tangkong:
If there are no further comments I'll merge this at end of day (8/25/22).  The disconnected devices issue will be a bit involved, and deserves a separate PR.

ZLLentz:
I agree, this is a strong improvement and fixes a lot!

PR #153, opened 2022-08-25
DOC: rework documentation

tangkong:
## Description
Rewrites documentation to reflect recent changes
updates doctr deploy keys for repository

## Motivation and Context
Docs were out of date.  Closes #140 

## How Has This Been Tested?
I build the docs a lot locally.  Hopefully travis can as well

## Where Has This Been Documented?
This IS documentation

PR #155, opened 2022-08-31
ENH: better handling of disconnected PV's, add config functionality

### LINKED ISSUE #109: Make LCLS-Specific Assumptions Configurable
## Expected Behavior

Any light source should be able to use lightpath, provided they know how their devices should be ordered and they satisfy the API

## Current Behavior

Required to use happi that sorts on z, beamline

## Possible Solution

Lightpath config file that picks which happi parameter to use for sorting sections, and which happi parameter to use for deciding what each section is

## Context

It's not simple to use this outside of LCLS for collaborative efforts

### LINKED ISSUE #152: BUG: Handle unreachable PV's / devices better
## Expected Behavior
We should be able to configure the EpicsSignal default timeout at startup, or rely on an environment variable

## Current Behavior
Currently we set the default EpicsSignal timeout to 10s, which worked around various `TimeoutErrors` and `pyepics failed to send connection and access rights`.  This works fine when all the devices are online, but devices are actually disconnected, this can cause large slowdowns when starting up lightpath (or when switching beamlines).  

Furthermore, switching to a beamline with disconnected devices seems to freeze the ui, as all disconnected devices time out at one of these function (`find_device_state`) calls in series before any device is shown in lightpath.

https://github.com/pcdshub/lightpath/blob/05a9394e09acf256649738f6233d54a7ba2b1b4d/lightpath/path.py#L83

## Possible Solution
Look for the timeout setting in an env variable, or add a cli argument to specify it.

Also handle disconnected devices more gracefully.  

## Context
This was first encountered in #142, and brought up again in #150 while testing during a mini-shutdown against a switch that was powered off, disabling a large number of devices. 

## Your Environment
During testing of #150

tangkong:
## Description
- Checks for PV connections before calling `get_lightpath_state`, skipping the timeouts
- Adds config functionality to cli and LightController, allowing the following to be set:
  - database
  - hutches to be loaded
  - EpicsSignal Timeout
  - minimum transmission
  - beamline config
  - source branch list
- sneaks in Pre-release notes from #153 because I forgot

## Motivation and Context
closes #109 
closes #152 

It bears noting that lightpath doesn't necessarily care if devices are disconnected.  It will represent them all the same.  Currently a huge number of devices start off as disconnected, but become responsive by the time the GUI opens.  This results in a lot of warning log messages spamming the terminal at startup.  I decided to mute these messages since disconnected status should be obvious via the GUI.

## How Has This Been Tested?
interactively, test has been added to the test suite

This PR.  If we're lucky the docs will pick up the cli changes automatically...


tangkong:
Moving the cli tests to the end of the test suite is a bandaid fix for now.  It's possible there are problems more systemic in lightpath, but I haven't been able to pin them down further.

Essentially, the cli tests cause a segfault in later tests.  If we only run one of the existing cli tests in the file (it doesn't matter which one), the suite will pass.  Running the cli tests in isolation or at the end of the suite also passes.  The even more strange part is that the segfault seems to arise based on the number of cli tests (duplicating any one of the tests will cause later gui tests to segfault).

A hand wavy explanation is that the cli tests create windows that don't get cleaned up properly by qtbot, and eventually qt or pytest runs out of memory.

I've chosen to move the tests to the end of this PR, but I'll leave a comment describing the danger at hand.

ZLLentz (commented):
I was planning to finish a review of this today but got sidetracked

lightpath/main.py
```diff
@@ -37,24 +54,38 @@ def main(db, hutches):
...
         Path to happi JSON database
     """
+    if cfg:
+        logger.info(f'reading config from: {cfg}...')
+        with open(cfg, 'r') as f:
+            conf = yaml.safe_load(f)
+    else:
+        conf = {}
 
+    timeout = conf.get('timeout', 10)  # timeout (s)
```
klauer, line 64:
```suggestion
    timeout = float(conf.get('timeout', 10))  # timeout (s)
```
Better to bail early here if the config is invalid

lightpath/main.py
```diff
@@ -25,10 +27,25 @@ def create_arg_parser():
...
+        if QApplication.instance() is None:
+            logger.debug("Creating QApplication ...")
+            qapp = QApplication([])
+        else:
+            logger.debug("Using existing QApplication")
+            qapp = QApplication.instance()
+    return qapp
+
+
+def main(db, hutches, cfg):
```
klauer, line 48: Docstring needs updating. Should `db`, `hutches` become optional now that both can be loaded from the config file?

Why? Well, what if I want to do `import lightpath.main; lightpath.main.main(cfg='...')`

lightpath/path.py
```diff
@@ -114,14 +114,23 @@ def find_device_state(device: Device) -> Tuple[DeviceState, LightpathState]:
     """
     # Gather device information
     try:
+        for sig in device.lightpath_summary._signals:
+            if not sig.connected:
+                # Check if relevant signals are connected
+                logger.debug(f"Unable to connect to device: {device}, "
```
klauer, line 120:
```suggestion
                logger.debug(f"Unable to connect to device: {device.name}, "
```
Device reprs can be annoyingly long

lightpath/tests/conftest.py
```diff
@@ -172,3 +175,24 @@ def simulated_lcls():
...
+    return cfg
+
+
+@contextmanager
+def cli_args(args):
+    """
+    Context manager for running a block of code with a specific set of
+    command-line arguments.
+    """
+    prev_args = sys.argv
```
klauer, line 195: Unnecessary to change as the end result is the same, but fyi - `parse_args` takes (I think) an `args=None` parameter, which can be a replacement for looking at `sys.argv`. Then your entrypoint would turn into `entrypoint(args=None)`. It's a bit cleaner than monkeypatching.

klauer (commented):
Looking good 👍 With that one fix Zach mentioned, I think this should be good to go.

ZLLentz:
Didn't see anything else when I reviewed the rest today- I'm :+1: after minor tweaks as above

lightpath/main.py
tangkong: On the other hand, cfg is optional if db and hutches are specified.  Requiring one of two argument (sets) in python is a bit weird, but I gave it a shot with `typing.overload`

lightpath/path.py
```diff
@@ -114,14 +114,23 @@ def find_device_state(device: Device) -> Tuple[DeviceState, LightpathState]:
     """
     # Gather device information
     try:
+        for sig in device.lightpath_summary._signals:
```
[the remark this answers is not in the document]
tangkong: Turns out this uncovered some weirdness in the mock devices.  The `SUB_META` callbacks don't properly mark their AggregateSignalState as connected, and need to be run again.  I don't quite understand why, but I've patched it for the simulated devices here.

ZLLentz (approved):

klauer (approved):

PR #157, opened 2022-09-14
PERF: remove hinted signals from LightRow widgets

tangkong:
## Description
- Removes hinted signals from lightrow widgets for performance reasons
- Removes cli check requiring one of `db` or `hutches`, defaulting to all the hutches in the default config
- Removes the `command` layout, allowing device widget centering
- ?Fixes? tests. Callbacks are being weird, nothing was changed within this PR that should have affected the stylesheet test, but the command hint test was rightfully removed.

## Motivation and Context
In the quest for a faster lightpath, we found that walking and creating the signal widgets on the main screen incurred a significant time loss.  The detailed screens still exist for allowing users to control their devices from lightpath.

As measured from the command line (`time python -m lightpath --hutches XCS`)
Before this PR: ~90s 
After this PR: ~30s 

This also lets us benefit from Zach's ophyd performance PR, since we are no longer waking up each device at lightpath startup.  

This does come with the caveat that opening detailed screens for the first time will be significantly slower.  We are essentially deferring the acquisition of that resource until later.

## How Has This Been Tested?
interactively

This PR

image

### DIFF
--- a/docs/source/upcoming_release_notes/157-perf_no_hinted_signals.rst
+++ b/docs/source/upcoming_release_notes/157-perf_no_hinted_signals.rst
@@ -1,0 +1,23 @@
+157 perf_no_hinted_signals
+##########################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- removes hinted signal widgets loaded from typhos to improve loadtimes
+- allows lightpath cli command to be run without arguments, loading all default hutches
+
+Contributors
+------------
+- tangkong
--- a/docs/source/upcoming_release_notes/157-perf_no_hinted_signals.rst
+++ b/docs/source/upcoming_release_notes/157-perf_no_hinted_signals.rst
@@ -15,7 +15,7 @@ Bugfixes
 
 Maintenance
 -----------
-- removes hinted signal widgets loaded from typhos to improve loadtimes
+- Removes hinted signal widgets loaded from typhos to improve loadtimes
 - allows lightpath cli command to be run without arguments, loading all default hutches
 
 Contributors
--- a/lightpath/ui/widgets.py
+++ b/lightpath/ui/widgets.py
@@ -89,9 +89,8 @@ class InactiveRow(Display):
 
     def condense(self):
         """Reduce the size of the widget when the device is hidden"""
-        # Hide commands and labels
+        # Hide and labels
         self.device_information.hide()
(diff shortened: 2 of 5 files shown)


tangkong:
CI failed once but passed on a re-run.  Hoping that tagging typhos will make that more consistent.

lightpath/ui/widgets.py
```diff
@@ -91,9 +89,8 @@ def clear_sub(self):
 
     def condense(self):
         """Reduce the size of the widget when the device is hidden"""
-        # Hide commands and labels
+        # Hide and labels
```
ZLLentz, line 92:
```suggestion
        # Hide labels
```

tangkong: 😞

ZLLentz (approved):
Performance considerations aside, I like the consistency this brings to screen. Previously there was no guarantee than any particular device hint buttons or readbacks would be particularly useful, though some of them may have been (slit gap size for example).

Perhaps at a later date if we find a way to speed it up this could return in some form, I did like seeing the state name for context for example, even if only to audit that the lightpath state made sense.

I wonder if there is any metadata we could slide into the `LightpathState` dataclass to use in this space in an efficient way instead of the ad hoc "grab a few signals and make widgets".

ZLLentz:
Just as an idea for a follow-up for what to do with the space, what if we extend like:
```
@dataclass
class LightpathState:
    inserted: bool
    removed: bool
    transmission: float
    output_branch: str
    state_summary: str
```
And then devices can provide text to lightpath to include in a label. Just a small idea, there are probably better ways to do this.

docs/source/upcoming_release_notes/157-perf_no_hinted_signals.rst
```diff
@@ -0,0 +1,23 @@
...
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- removes hinted signal widgets loaded from typhos to improve loadtimes
```
klauer, line 18:
```suggestion
- Removes hinted signal widgets loaded from typhos to improve loadtimes
```

Total nitpick, but the docs will look much nicer if these are capitalized appropriately

klauer (approved):
I saw this in person and liked it. 👍 

The code here is almost exclusively removing things, so there's not much to say beyond that...

klauer:
`state_summary` sounds nice - agreed about that label.

tangkong:
I'll toss the `state_summary` idea into an issue.  👍

Issue #159, opened 2022-09-19
GUI: Add refresh button to screen (read: update the active path when it changes)


tangkong:
## Expected Behavior
There should be a way to refresh the devices and update the path being used. A refresh button would be the simplest way to do this, though we could be more sneaky and poll the controller to see if the active path has changed.  

Say if we start in a configuration that routes beam to XCS via L0 -> XRT -> XCS_LODCM -> XCS.  If we insert mr1l3 in a way that diverts beam to the L3 path, the lightpath gui will not change the devices in the beam.  

![image](https://user-images.githubusercontent.com/35379409/191128677-2db680b1-26ac-462a-9b5c-b0290eb409b5.png)

## Current Behavior

There is no refresh button at the moment.  If the active path changes during operation, we must switch off the current beamline and back.  This works, and isn't all that slow thanks to recent speedups.  But it's annoying, and possibly unintuitive

There is also no guarantee that the user will have multiple beamlines loaded, in which case the user will not be able to switch active paths. 

## Possible Solution

Refresh button?
add a `LightController.active_path` call to the callback?  (this might be a bit too much given the rate of signal updates)

## Context
With new LODCM multi-output functionality, I happened upon this issue.  It's always been a problem but is even more prominent with the possibility of multiple beam destinations

## Your Environment
pcds-5.4.2, lightpath master

tangkong:
closed by #162

Issue #136, opened 2022-07-12
BUG: Icon clipping at top of GUI


tangkong:
## Expected Behavior
I see a bit of icon clipping in the summary at the top of the GUI:
<img width="269" alt="image" src="https://user-images.githubusercontent.com/5139267/178569869-b957b48b-d422-49a5-9d03-61a0fe46e4d1.png">

## Context
_Originally posted by @klauer in https://github.com/pcdshub/lightpath/pull/133#pullrequestreview-1036074786_

## Your Environment
pcds-5.4.1 + #133

tangkong:
This was tweaked a bit in #162, while not perfect it's much better
<img width="341" alt="image" src="https://user-images.githubusercontent.com/35379409/195443060-17ec22c6-b7ac-4e75-a43d-0e4da4c67b23.png">

Issue #82, opened 2018-08-24
Loading Screen


teddyrendahl:
## Expected Behavior

The `BeamPath` can take a long time to load. It might be nice if we have a progress bar that comes up to let the operator know something is happening.

tangkong:
closed by #162

PR #162, opened 2022-09-22
GUI: add loading screen and refresh button

tangkong:
## Description
- adds a simple splash screen that pops up during `change_path_display()` calls, closing #82 
- adds a refresh button, closing #159 
- hopefully fixes clipping in overview slider with condensed lightrow widgets, closing #136 

## Motivation and Context
Linked issues above.  The splash screen was shamelessly copied and modified from lucid.

I'm looking into also possibly making the detailed screen width adjustable, since it takes up a lot of room as is.  But that's for tomorrow possibly.

## How Has This Been Tested?
Both sim and real-device lightpaths have been clicked around.  

This PR text

image

image

### DIFF
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -82,6 +82,8 @@ class LightApp(Display):
         self.upstream_device_combo.activated[str].connect(self.update_upstream)
         self.remove_check.toggled.connect(self.filter)
         self.detail_hide.clicked.connect(self.hide_detailed)
+        self.refresh_button.clicked.connect(self.change_path_display)
+
         # Store LightRow objects to manage subscriptions
         self.rows = list()
         # store device type filter widgets
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -1,6 +1,7 @@
 """
 Full Application for Lightpath
 """
+import contextlib
 import logging
 import os.path
 import threading
@@ -106,9 +107,10 @@ class LightApp(Display):
 
     def destinations(self):
         """
-        All possible beamline destinations
+        All possible beamline destinations that have an associated path
         """
-        return list(self.light.beamlines.keys())
+        return [line for line in self.light.beamlines.keys()
+                if self.light.beamlines[line]]
 
     def load_device_row(self, device):
         """
@@ -200,12 +202,7 @@ class LightApp(Display):
         """
         Change the display devices based on the state of the control buttons
         """
-        self.loading_splash.move(self.geometry().center()
-                                 - self.loading_splash.rect().center())
-        self.loading_splash.show()
-        self.loading_splash.update_status(f'{self.selected_beamline()} path')
-
-        with self._lock:
+        with self._lock, self.open_splash(f'{self.selected_beamline()} path'):
             logger.debug("Resorting beampath display ...")
             # Remove old detailed screen
             self.hide_detailed()
@@ -257,7 +254,27 @@ class LightApp(Display):
         # Update device type checkboxes
         self.update_device_types()
         self.setWindowTitle(f'Lightpath - {self.selected_beamline()}')
-        self.loading_splash.hide()
+
+    @contextlib.contextmanager
+    def open_splash(self, msg: str):
+        """
+        Context manager for opening the loading splash screen, and
+        closing it always
+
+        Parameters
+        ----------
+        msg : str
+            message to show in the loading screen
+        """
+        try:
+            self.loading_splash.move(self.geometry().center()
+                                     - self.loading_splash.rect().center())
+            self.loading_splash.show()
+            self.loading_splash.update_status(msg)
+            yield
+
+        finally:
+            self.loading_splash.hide()
 
     def ui_filename(self):
         """
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -48,7 +48,7 @@ class LightApp(Display):
                  parent=None, dark=True):
         super().__init__(parent=parent)
         # Store Lightpath information
-        self.loading_splash = LoadingSplash()
+        self.loading_splash = LoadingSplash(parent=self)
         self.light = controller
         self.path = None
         self.detail_screen = None
--- a/lightpath/ui/gui.py
+++ b/lightpath/ui/gui.py
@@ -11,7 +11,8 @@ import typhos
 from pydm import Display
 from qtpy.QtCore import Qt
 from qtpy.QtCore import Slot as pyqtSlot
-from qtpy.QtWidgets import QCheckBox, QGridLayout, QHBoxLayout
+from qtpy.QtWidgets import (QApplication, QCheckBox, QDialog, QGridLayout,
+                            QHBoxLayout, QLabel, QVBoxLayout)
 from typhos import TyphosDeviceDisplay
 
 from lightpath.path import DeviceState
@@ -46,6 +47,7 @@ class LightApp(Display):
                  parent=None, dark=True):
         super().__init__(parent=parent)
         # Store Lightpath information
+        self.loading_splash = LoadingSplash()
         self.light = controller
         self.path = None
         self.detail_screen = None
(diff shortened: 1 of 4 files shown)


lightpath/ui/gui.py
```diff
@@ -411,3 +421,33 @@ def resizeEvent(self, evt):
...
+        status_layout = QHBoxLayout()
+        status_layout.addWidget(self.status_display)
+        status_layout.addWidget(loading)
+
+        layout.addLayout(status_layout)
+
+    def update_status(self, msg):
+        self.status_display.setText(f"Loading: {msg}")
+        # self.move()
+        QApplication.instance().processEvents()
```
ZLLentz, line 453: I think this is the only bit I don't understand- why do we need to explicitly process events here?

lightpath/ui/gui.py
```diff
@@ -411,3 +421,33 @@ def resizeEvent(self, evt):
...
+
+        status_layout = QHBoxLayout()
+        status_layout.addWidget(self.status_display)
+        status_layout.addWidget(loading)
+
+        layout.addLayout(status_layout)
+
+    def update_status(self, msg):
+        self.status_display.setText(f"Loading: {msg}")
+        # self.move()
```
ZLLentz, line 452:
```suggestion
```

ZLLentz:
I was hoping to run it myself too but something is wrong with my dev setup. I need to grab some food and then I'll try again.

lightpath/ui/gui.py
tangkong: If you don't process events, the QDialog will appear but not render any of the text or internals.  I'm not entirely sure why, but I think it has to do with processing the events in that thread, and not letting `change_path_display` block.

ZLLentz: Is this also why the animation doesn't play when we switch subdisplays?

lightpath/ui/gui.py
```diff
@@ -248,6 +257,7 @@ def change_path_display(self, value=None):
         # Update device type checkboxes
         self.update_device_types()
         self.setWindowTitle(f'Lightpath - {self.selected_beamline()}')
+        self.loading_splash.hide()
```
ZLLentz, line 260: if you select `TXI`, it ends up finding zero devices and the splash loading display gets stuck open forever

lightpath/ui/gui.py
```diff
@@ -196,6 +200,11 @@ def change_path_display(self, value=None):
         """
         Change the display devices based on the state of the control buttons
         """
+        self.loading_splash.move(self.geometry().center()
+                                 - self.loading_splash.rect().center())
```
ZLLentz, line 204: These coordinates aren't correct for me on first load, the initial splash appears awkwardly at the top of one of my monitors in a half-loaded/frozen state

lightpath/ui/gui.py
tangkong: This is a good catch, we can both handle the opening/closing of the screen better and also not give users the option to open a beamline with no devices/beampath.  Addressed in df288b0

lightpath/ui/gui.py
tangkong: I believe so.  I haven't been able to find a way to get the animation to run, but I thought it might be ok to leave as is for now.

lightpath/ui/gui.py
```diff
@@ -196,7 +202,7 @@ def change_path_display(self, value=None):
         """
         Change the display devices based on the state of the control buttons
         """
-        with self._lock:
+        with self._lock, self.open_splash(f'{self.selected_beamline()} path'):
```
klauer, line 205: I had to double-check the guarantees Python gives for multiple `with` expressions:
image (https://docs.python.org/3/reference/compound_stmts.html)
This should be fine, then.

Thinking more about the `threading.Lock` mechanism here:
* A lock should be around a section where you don't want other threads to execute the same code at the same time
* This necessarily has to be called in the GUI thread (because it interacts with widgets)
* If somehow `change_path_display` gets called during this block, it'll cause a deadlock
* Is there some other scenario I'm not seeing?

So... I guess I'm glad the weird syntax made me think more about this code block. 
If it works fine as-is, it's probably OK to leave it...

lightpath/ui/gui.py
klauer: I think it's as you guessed - you're tying up the GUI thread in `change_path_display` and not allowing Qt to process events for other events happening in the background (or even foreground redraws).

lightpath/ui/widgets.py
```diff
@@ -233,6 +234,7 @@ def __init__(self, device, parent=None):
         # Default UI settings for conformity
         self.setMinimumSize(10, 10)
         self.setMaximumSize(50, 50)
+        self.setStyleSheet('padding : 0px')
```
klauer, line 237: Totally optional tangent: 
I had started putting stylesheet stuff in a separate file in btms-ui. 
I think it worked reasonably well:
* https://github.com/pcdshub/btms-ui/blob/master/btms_ui/stylesheet.qss
* https://github.com/pcdshub/btms-ui/blob/c6c6d7aaf95fc39bcd2ac754e1cd01c18af2d82b/btms_ui/main.py#L21

You might consider something similar down the line. It's nice to have all style stuff in one place.

klauer (approved):
It worked for me 👍 

Startup:
image

Switching:
image

tangkong:
I'll give Zach another chance to take a look, but I plan to merge this tomorrow so I can tag and start fighting the integration battles.  🤞

lightpath/ui/gui.py
```diff
@@ -1,6 +1,7 @@
 """
```
ZLLentz, line 1: No idea where it comes from, but lightpath prints this at the end of execution:
```
<lightpath.ui.gui.LightApp object at 0x7fb92a00a280>
```

lightpath/ui/gui.py
ZLLentz: It either doesn't appear at all (most common) or appears rendering oddly at the top of where the screen eventually appears. I tried to get a screenshot but it's rare and I gave up trying to repeatedly open it.

ZLLentz (approved):
Should be good

lightpath/ui/gui.py
tangkong: Following a brilliant suggestion from zach, I just replaced the animation with a static loading icon.

lightpath/ui/gui.py
klauer: I think command-line entrypoints by default print the result of the function if not `None`.
We should probably remove the `return` from `return main(...)` in the entrypoint:
https://github.com/pcdshub/lightpath/blob/09be87d335952d39a597f2fe8bb1576bd8f2ece3/lightpath/main.py#L130

ZLLentz: I'm using pcds-5.4.2 + PYTHONPATH and running via `lightpath`.
If you don't get this print and the entrypoint is still wonky on that env then let's disregard this for now.

klauer: This has puzzled me for a while, where exactly that printout came from. There is no obvious spot in the entrypoint mechanism. I believe this explains it:
```
$ python
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:52)
[Clang 13.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.exit("puzzling printout")
puzzling printout
```

More tersely:
```
$ python -c "import sys; sys.exit('test')"; echo $?
test
1
```

And the `lightpath` script has:
```python
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(load_entry_point('lightpath', 'console_scripts', 'lightpath')())
```

Thus the return value from the entrypoint gets set as the argument to `sys.exit`. If it's not numeric, its string representation gets dumped to standard output and the process returns exit code `1`.

Removing the `return` from the entrypoint function should resolve this.

ZLLentz: that... is really bizarre and makes a lot of sense

tangkong: We should add this to the cookie cutter or something, I feel like several packages' entrypoints have needed Ken's intervention recently.  Nevertheless I'll toss an issue up for this 👍

PR #163, opened 2022-09-28
DOC: release notes for v1.0.0

tangkong:
## Description
Release notes for v1.0.0

## Motivation and Context
I hope I did this right...

## How Has This Been Tested?
I ran `make html` a few times locally

✨ This is documentation ✨


ZLLentz:
Months of hard work summarized in a single page

ZLLentz:
I don't think I can possible review for completeness but let's check for typos and such

docs/source/releases.rst
```diff
@@ -0,0 +1,79 @@
...
+###############
+
+v1.0.0 (2022-09-28)
+===================
+
+API Changes
+-----------
+- LightpathState combines transmission and output_branch into a single
+  field called output, a mapping from output_branch to transmission on
+  that branch
```
ZLLentz, line 11: I think this point is an API change between two non-tagged versions, I think that means it can be skipped here? E.g. `LightpathState` didn't exist at all in the previous tag

docs/source/releases.rst
```diff
@@ -0,0 +1,79 @@
...
+- Extends the default EpicsSignal timeout to allow devices to connect
+- Prevent beamlines with no beampath from destination combo box
+- Removes endstation sorting to allow lazy BeamPath loading
+- Fixes various font-awesome deprecation warnings
+
+Maintenance
+-----------
+- GUI now subscribes to lightpath_summary signals instead of the device
+- Remove unneeded calls to ``lightpath_summary.get``, previously used
+  to initialize ``lightpath_summary``
```
ZLLentz, line 60: I think this is another between-dev-versions fix

ZLLentz (approved):
Big tag

docs/source/releases.rst
tangkong: True, I'll fix that up

klauer:
👏

PR #167, opened 2022-10-21
BUG: read the updated configuration when calling load_beamlines

tangkong:
## Description
`LightController.load_beamlines` was mistakenly still referencing the built-in `beamlines` config/dictionary, despite the config being updated at `__init__`.  

https://github.com/pcdshub/lightpath/blob/09be87d335952d39a597f2fe8bb1576bd8f2ece3/lightpath/controller.py#L147

This did not show up as a bug because subsequent paths to `LightController.get_paths` did this correctly, and as long as the graph was initialized (using `load_beamlines`) to have the required paths, `get_paths` would find the path and split it according to the new (provided) cfg

https://github.com/pcdshub/lightpath/blob/09be87d335952d39a597f2fe8bb1576bd8f2ece3/lightpath/controller.py#L192

https://github.com/pcdshub/lightpath/blob/09be87d335952d39a597f2fe8bb1576bd8f2ece3/lightpath/controller.py#L212

Where the bug appears is if you create a config with beamlines/sources not present in the default config.  In this case, the graph would not be initialized correctly, and the `get_paths` call would fail to find a path through the graph

(Also slips in an actual error message for when there are no devices found in a branch)

## Motivation and Context
In fixing up the hutch-python usage of lightpath, I tried testing with arbitrary beamlines/branch names and stumbled upon this.

## How Has This Been Tested?
Interactively.  
I should add a test.

This PR

### DIFF
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -144,7 +144,7 @@ class LightController:
             Name of endstation to load
         """
         try:
-            end_branches = beamlines[endstation]
+            end_branches = self.cfg['beamlines'][endstation]
         except KeyError:
             logger.warning("Unable to find %s as a configured endstation, "
                            "assuming this is an invalid path", endstation)
@@ -183,6 +183,7 @@ class LightController:
         List[BeamPath]
             a list of BeamPath's to the requested endstation
         """
+        # if path exists, return it
         paths = self.beamlines[endstation]
 
         if all([isinstance(path, BeamPath) for path in paths]):
--- a/lightpath/controller.py
+++ b/lightpath/controller.py
@@ -265,6 +265,9 @@ class LightController:
             the active path
         """
         paths = self.get_paths(dest)
+        if len(paths) == 0:
+            raise PathError('No paths in facility to the '
+                            f'desired endstation: {dest}')
         if len(paths) == 1:
             return paths[0]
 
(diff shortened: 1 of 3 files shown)


tangkong:
I fixed the tests but conda-forge is failing due to timeouts.  This is a next week problem

tangkong:
I just re-ran tests until the conda build actually progressed... and things finally ran correctly.  🤷

klauer (approved):
This is a good bugfix and looks fine to me.

Likely annoying aside: it does make me wonder if we should be dealing with the `cfg` dictionary after `__init__` at all, though. Naively, I'd expect things like `self.cfg["beamlines"]` to just be `self.beamlines`. Is there a benefit to managing that dictionary instead of making its contents first-class attributes?

tangkong:
pulling the cfg dictionary apart is probably wise.  This issue was spawned by the confusion between
- `LightController.beamlines`: which holds the BeamPath objects or unfilled node lists, and
- `LightController.cfg.beamlines`, which holds the configuration information for the facility.  

I'll refactor that here, ultimately we shouldn't need to tag this to get the other packages that rely on lightpath (hutch_python) running in the new env

klauer:
Additional changes look good 👍 

 I'd say may as well merge and tag if Zach is going to re-run the environment building script anyway

PR #168, opened 2022-10-25
DOC: release notes for v1.0.1

tangkong:
## Description
Quick bugfix tag

## Motivation and Context
get a bugfix out before the new env?

## How Has This Been Tested?
`make html` ran without issue

This is documentation yay

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-pytest
+pytest<7.2.0
 doctr
 ipython
 matplotlib
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -9,6 +9,10 @@ Bugfixes
 - ``LightController.load_beamlines`` now properly references the provided
   configuration if provided, rather than the default ``beamlines`` config
 
+Contributors
+------------
+- tangkong
+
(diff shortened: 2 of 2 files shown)


ZLLentz (approved):
LGTM

You can add a contributors section if you want and slot your name in

ZLLentz:
What happened to the travis build though? It didn't even get to the running-our-code part. I'll restart that job.

ZLLentz:
The tests passed on the previous PR so this must be a weird CI issue- maybe some random dependency or test dependency updated? I suspect we can still run `pytest` interactively without issues.

ZLLentz:
`pytest` itself got an update on `conda-forge` like 2 hours ago, it could be related

ZLLentz:
Possibly a packaging issue related to:
```
pytest no longer depends on the py library. pytest provides a vendored copy of py.error and py.path modules but will use the py library if it is installed. If you need other py.* modules, continue to install the deprecated py library separately, otherwise it can usually be removed as a dependency.
```

tangkong:
Ok with a pytest pin, these tests pass.  I'm going to merge and tag

PR #170, opened 2022-11-15
MAINT: scripted fix for precommit

ZLLentz:
PR generated from script to fix .pre-commit-config.yaml in all repos, switching out the missing flake8 gitlab mirror for github.


tangkong (approved):

tangkong:
I think it's fine, probably?  I feel like I wouldn't be annoyed if this change missed the release notes..

ZLLentz:
You're probably right but I guess I just want to do it all the way

Issue #169, opened 2022-10-27
Intermittent test failures from pcds-envs


ZLLentz:
## Expected Behavior

Tests should always pass!

## Current Behavior

See https://github.com/pcdshub/pcds-envs/issues/253
<details>

`lightpath/tests/test_gui.py::test_upstream_check` times out occasionally (10 minutes)

```
________________________________ test_filtering ________________________________
lightapp = <lightpath.ui.gui.LightApp object at 0x7faf975e80d0>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7faf9c1b37f0>
    def test_filtering(lightapp: LightApp, monkeypatch):
        lightapp.destination_combo.setCurrentIndex(4)  # set current to MEC
        # Create mock functions
        for row in lightapp.rows:
            monkeypatch.setattr(row[0], 'setHidden', Mock())
        # Initialize properly with nothing hidden
        lightapp.filter()
        for row in lightapp.rows:
            row[0].setHidden.assert_called_with(False)
        # Insert at least one device then hide
        device_row = lightapp.rows[2][0]
        device_row.device.insert()
        lightapp.remove_check.setChecked(False)
        # Reset mock
        for row in lightapp.rows:
            row[0].setHidden.reset_mock()
        lightapp.filter()
        for row in lightapp.rows:
            if row[0].device.get_lightpath_state().removed:
>               row[0].setHidden.assert_called_with(True)
lightpath/tests/test_gui.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <Mock id='140392215288080'>, args = (True,), kwargs = {}
expected = call(True), actual = call(False)
_error_message = <function NonCallableMock.assert_called_with.<locals>._error_message at 0x7faf9762c9d0>
cause = None
    def assert_called_with(self, /, *args, **kwargs):
        """assert that the last call was made with the specified arguments.
    
        Raises an AssertionError if the args and keyword args passed in are
        different to the last call to the mock."""
        if self.call_args is None:
            expected = self._format_mock_call_signature(args, kwargs)
            actual = 'not called.'
            error_message = ('expected call not found.\nExpected: %s\nActual: %s'
                    % (expected, actual))
            raise AssertionError(error_message)
    
        def _error_message():
            msg = self._format_mock_failure_message(args, kwargs)
[...]

PR #175, opened 2023-01-13
WIP/BLD: GHA migration

tangkong:
## Description
Runs `pcds-python-migration-tools:update_repository.py` from https://github.com/pcdshub/pcds-python-migration-tools/pull/3

## Motivation and Context
Trying out GHA with Ken's spiffy new migration tools

## How Has This Been Tested?
testing here?

documentation here?


klauer:
For those conda build failures, nuke your `build.sh` and fix `meta.yaml` to have these lines:

https://github.com/pcdshub/cookiecutter-pcds-python/blob/72f80d77e5bf13d2e5b4a3d0343fde4685688bda/%7B%7B%20cookiecutter.folder_name%20%7D%7D/conda-recipe/meta.yaml#L1-L16

I think I want to add that to the migration tool, too...

klauer:
I used your notes to fix up the migration tool a bit more - thanks!

One remaining thing: add `setuptools_scm` and `pip` as build deps in the conda recipe

Rest is probably good 🤞

tangkong:
3.11 Failures:
- conda fails on build, a compatible version of scipy not available I suppose
- pip builds but tests fail.  [One of the failures](https://github.com/pcdshub/lightpath/actions/runs/3915469285/jobs/6693673747#step:14:2378) seems to be of the intermittent kind, as seen during the [env build](https://github.com/pcdshub/lightpath/issues/169).  The [other](https://github.com/pcdshub/lightpath/actions/runs/3915469285/jobs/6693673747#step:14:2294) is a strange typhos issue where the panel `kinds` aren't being initialized properly?

tangkong:
Although the 3.11 tests are marked as experimental and do not prevent the overal test job from failing, the little status icon next to the commit hash still suggests failure. :/

It could just be something to get used to, but I didn't uncover anything during my brief search

image

klauer:
> Although the 3.11 tests are marked as experimental and do not prevent the overal test job from failing, the little status icon next to the commit hash still suggests failure. :/

This needs to be configured here in the branch protection settings:

image

It's sensible from GitHub's standpoint: users might want to have different rules depending on the branch being merged into. And that shouldn't be baked into the actions configuration itself.

Since we want a generally consistent set of settings., I'm not sure if there's a better or easier way for us to configure each repository.

klauer:
For 3.11's test failure, it's not typhos's fault entirely:

```python
In [1]: from ophyd.ophydobj import Kind

In [2]: list(Kind)
Out[2]: [<Kind.normal: 1>, <Kind.config: 2>]
```

Looks to be an ophyd bug or incompatibility with 3.11's changes

tangkong:
I suppose I expected the `experimental` setting to affect this, but in our case since `fail_fast` is false it doesn't affect the workflow.  

Having to manually set up each repository's check failure settings is annoying, and adding another moving piece to update when our ci helpers change sounds bad to me.

klauer:
> Having to manually set up each repository's check failure settings is annoying, and adding another moving piece to update when our ci helpers change sounds bad to me.

Completely agreed.

I've found that the GitHub API can be used to list/add/delete branch protection rules, meaning we could automate the procedure. If that ends up not working well, I think it'll be time to re-evaluate the job matrix.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[tool.setuptools_scm]
+write_to = "lightpath/_version.py"
+
+[project.entry-points."happi.containers"]
+lightpath = "lightpath.happi.containers"
+
+[tool.setuptools.packages.find]
+where = [ ".",]
+include = [ "lightpath",]
```
klauer, line 33:
```suggestion
include = [ "lightpath*",]
```

Should fix https://github.com/pcdshub/lightpath/actions/runs/3941423539/jobs/6743810335#step:11:140

setup.py
```diff
@@ -1,31 +0,0 @@
...
-
-setup(
-    name="lightpath",
-    version=versioneer.get_version(),
-    cmdclass=versioneer.get_cmdclass(),
-    license="BSD",
-    author="SLAC National Accelerator Laboratory",
-    packages=find_packages(),
-    package_data={
-        "lightpath": ["ui/*.ui"],
```
klauer, line 20: Double-checking that `ui/*.ui` gets included in the source distribution and wheel:
```
opying lightpath/ui/__init__.py -> lightpath-1.0.2.dev36+g22b74e9/lightpath/ui
copying lightpath/ui/device.ui -> lightpath-1.0.2.dev36+g22b74e9/lightpath/ui
copying lightpath/ui/gui.py -> lightpath-1.0.2.dev36+g22b74e9/lightpath/ui
copying lightpath/ui/lightapp.ui -> lightpath-1.0.2.dev36+g22b74e9/lightpath/ui
```

👍

pyproject.toml
tangkong: Ah good catch, I didn't step through all the CI steps thoroughly enough d'oh

.pre-commit-config.yaml
```diff
@@ -24,11 +23,11 @@ repos:
...
 
 -   repo: https://github.com/pycqa/flake8.git
-    rev: 3.9.2
+    rev: 6.0.0
     hooks:
     -   id: flake8
 
 -   repo: https://github.com/timothycrosley/isort
-    rev: 5.10.1
+    rev: 5.11.4
```
ZLLentz, line 31: Maybe it's a good idea to bump this again after our learning today

lightpath/version.py
```diff
@@ -0,0 +1,47 @@
+"""
```
ZLLentz, line 1: Do I still need to fix the bug in this boilerplate file that makes it miss the repo-level `.git_archival.txt`?

ZLLentz (approved):
This looks good to me. It might make sense to merge one of these to debug the "download-from-github-as-tarball" path on the version selection, but that might not actually ever be needed.

lightpath/version.py
klauer: I'm not sure if we converged on a proper solution for that... Was it the result of the "let's fix the version 0.0.0 shown when building the conda package?" or something else?

ZLLentz: I think the fix you applied to pcdsdevices is the correct one

klauer: Right, so correct me if I'm remembering incorrectly - the `0.0.0` fix was made by way of adding in setuptools-scm to the conda build requirements.

The addition of the `git_archival` clause in `version.py` came along at the same time in an attempt to fix that, but it also made us realize simultaneously that a release source code tarball on GitHub would not be properly versioned unless we modified that line.

I _think_ that's how it went. In that case, I agree that we should update the template and this `version.py` to include a check for `git_archival.txt`

lightpath/version.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Optional
+
+
+class VersionProxy(UserString):
+    def __init__(self):
+        self._version = None
+
+    def _get_version(self) -> Optional[str]:
+        # Checking for directory is faster than failing out of get_version
+        if (Path(__file__).parent.parent / '.git').exists():
```
klauer, line 18:
```suggestion
        if (repo_root / ".git").exists() or (repo_root / ".git_archival.txt").exists():
```

klauer (approved):

PR #176, opened 2023-02-04
MNT: split XPP into XPP-PINK and XPP-MONO

tangkong:
## Description
Splits XPP's single entry in the default config into two: `XPP_MONO` and `XPP_PINK`.  
Fixes the test suite accordingly
Fixes a small bug on the way that didn't re-filter the visible devices after changing a path.

## Motivation and Context
All caps because of some `string.upper()` stuff we do in the GUI.

Previously XPP would have two possible beampaths splitting at the xpp lodcm.  One going on the L0 branch and one on the L2 branch.  While before we were letting lightpath choose which branch is active (by whichever branch has the sees beam at the highest z), it became clear that XPP would want to see the status of either line independently.  

The test_path/test_callback test failed again, but passed after a re-run. I tried adding in a sleep clause to possibly let it process, but the failures are intermittent enough that it'll be tough to nail down.  This is not the only test that might benefit from this kind of pause, as many other tests invoke similar code paths.  

Perhaps we just resign ourselves to re-running test jobs.  At least GHA shows past results from re-run jobs

## How Has This Been Tested?
Interactively.

This PR, and [JIRA](https://jira.slac.stanford.edu/browse/LCLSECSD-1630)

### DIFF
--- a/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
+++ b/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
@@ -1,0 +1,22 @@
+176 mnt_split_xpp
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- filters the visible LightRow widgets after a path change. This fixes a bug where changing paths would show all devices on a path, ignoring the state of show-removed or device filter checkboxes.
+
+Maintenance
+-----------
+- Splits the XPP line into 'XPP_PINK' and 'XPP_MONO' so both are visible and selectable.  This schange only affects the default config file; the ability to specify multiple paths or endpoints for a single name in the config remains.
+
+Contributors
+------------
+- tangkong
--- a/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
+++ b/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
@@ -15,7 +15,7 @@ Bugfixes
 
 Maintenance
 -----------
-- Splits the XPP line into 'XPP_PINK' and 'XPP_MONO' so both are visible and selectable.  This schange only affects the default config file; the ability to specify multiple paths or endpoints for a single name in the config remains.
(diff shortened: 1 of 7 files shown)


lightpath/ui/gui.py
```diff
@@ -374,6 +376,10 @@ def filter(self, *args):
             # "upstream" devices now.  Possibly by branch name?
             hidden_upstream = (device.md.z < upstream_device_z)
             # Hide device if any of the criteria are met
+            if row[0].device.name == 'xcs_lodcm':
```
klauer, line 379: Debug prints to remove

docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
```diff
@@ -0,0 +1,22 @@
...
+--------
+- N/A
+
+Bugfixes
+--------
+- filters the visible LightRow widgets after a path change. This fixes a bug where changing paths would show all devices on a path, ignoring the state of show-removed or device filter checkboxes.
+
+Maintenance
+-----------
+- Splits the XPP line into 'XPP_PINK' and 'XPP_MONO' so both are visible and selectable.  This schange only affects the default config file; the ability to specify multiple paths or endpoints for a single name in the config remains.
```
klauer, line 18:
```suggestion
- Splits the XPP line into 'XPP_PINK' and 'XPP_MONO' so both are visible and selectable.  This change only affects the default config file; the ability to specify multiple paths or endpoints for a single name in the config remains.
```

klauer (approved):
GUI testing is notoriously finicky...

I'm totally fine with re-running the test suite until we get around to solving the underlying issue.

lightpath/ui/gui.py
tangkong: whooooops thanks

ZLLentz (approved):
Should be good!

ZLLentz:
The 3.11 conda build failure is probably due to the pyqt pin 
The other 3.11 failure may be harder to track down

PR #178, opened 2023-02-11
REV: revert split of xpp line

tangkong:
## Description
reverts the config split of XPP into two lines for hutch-python compatibility

## Motivation and Context
Hutch-python expects very specific hutch names, so we should not change the ones that are available by default. 

This split can simply be done via custom config files in the future

Can't just revert the whole PR since there were other fixes in it

## How Has This Been Tested?
reverting and running the test suite

This PR

### DIFF
--- a/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
+++ b/docs/source/upcoming_release_notes/176-mnt_split_xpp.rst
@@ -15,7 +15,6 @@ Bugfixes
 
 Maintenance
 -----------
-- Splits the XPP line into 'XPP_PINK' and 'XPP_MONO' so both are visible and selectable.  This change only affects the default config file; the ability to specify multiple paths or endpoints for a single name in the config remains.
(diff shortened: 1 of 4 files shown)


ZLLentz:
Another option is to teach hutch-python about the two XPP lines, I'm not sure which is better

ZLLentz (approved):
If you think this is the way to go, the diff looks correct (inverse of the functional bits of #176).
Do you need to revise the pre-release notes from that PR to remove this from it?

klauer (approved):
The reversion looks fine to me.

> Another option is to teach hutch-python about the two XPP lines, I'm not sure which is better

Would you mind giving a short explanation as to why the route chosen in this PR is better (?) than the other? Or is it just a quicker fix?

ZLLentz:
You're definitely right that this would end up as a special case in hutch-python and possibly have knock-on impacts

klauer:
Got it - this clearly seems the correct approach then 👍

PR #179, opened 2023-02-22
BLD: release pyqt pin

tangkong:
## Description
Try releasing the pyqt pin

## Motivation and Context
Maybe this fixes our matplotlib py3.11 issues?

## How Has This Been Tested?
may CI be graceful to us

This PR


klauer (approved):

tangkong:
This ends up not fixing our CI issues, but it's probably good to not carry the pin here if unneeded

PR #185, opened 2023-07-06
TST: wait for conditions during troublesome tests

tangkong:
## Description
Try to fix tests in #169.  Changes like these could be spread in many places

Basically just assert things with a timeout to give conditions the ability to settle.

## Motivation and Context
#169 , and other tests

## How Has This Been Tested?
Locally, and by hitting the GHA re-run button a bunch of times, likely

This PR.

### DIFF
--- a/lightpath/tests/test_gui.py
+++ b/lightpath/tests/test_gui.py
@@ -2,6 +2,7 @@ from distutils.spawn import find_executable
 from unittest.mock import Mock
 
 import pytest
+from pytestqt.qtbot import QtBot
 
 from lightpath.controller import LightController
 from lightpath.ui import LightApp
@@ -68,7 +69,7 @@ def test_upstream_check(lightapp: LightApp, monkeypatch):
             row[0].setHidden.assert_called_with(False)
 
 
-def test_filtering(lightapp: LightApp, monkeypatch):
+def test_filtering(qtbot: QtBot, lightapp: LightApp, monkeypatch):
     lightapp.destination_combo.setCurrentIndex(4)  # set current to MEC
     # Create mock functions
     for row in lightapp.rows:
@@ -76,7 +77,7 @@ def test_filtering(lightapp: LightApp, monkeypatch):
     # Initialize properly with nothing hidden
     lightapp.filter()
     for row in lightapp.rows:
-        row[0].setHidden.assert_called_with(False)
+        qtbot.waitUntil(lambda: row[0].setHidden.assert_called_with(False))
     # Insert at least one device then hide
     device_row = lightapp.rows[2][0]
     device_row.device.insert()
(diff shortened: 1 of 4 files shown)


tangkong:
I can expand this treatment to other tests, but I also haven't managed to see a test suite failure throughout the course of this PR.  If anyone has a better stress test methodology I'd love to try it

ZLLentz:
One way to stress test this is to use pytest-repeat but that might reveal other test suite issues
```
pytest --count=1000 tests/bad_tests.py
```

lightpath/tests/test_widgets.py
```diff
@@ -27,15 +28,32 @@ def test_widget_updates(lightrow: LightRow, path: BeamPath):
...
     ipimb.insert()
-    assert (to_stylesheet_color(state_colors['half_removed'])
-            in ipimb_row.state_label.styleSheet())
+
+    def half_removed():
+        return (to_stylesheet_color(state_colors['half_removed'])
+                in ipimb_row.state_label.styleSheet())
+
+    wait_until(half_removed)
+    assert half_removed
```
ZLLentz, line 37: I think these are all typos, you meant to assert the function result, right?
```suggestion
    assert half_removed()
```

lightpath/tests/test_gui.py
```diff
@@ -68,15 +69,15 @@ def test_upstream_check(lightapp: LightApp, monkeypatch):
...
+def test_filtering(qtbot: QtBot, lightapp: LightApp, monkeypatch):
     lightapp.destination_combo.setCurrentIndex(4)  # set current to MEC
     # Create mock functions
     for row in lightapp.rows:
         monkeypatch.setattr(row[0], 'setHidden', Mock())
     # Initialize properly with nothing hidden
     lightapp.filter()
     for row in lightapp.rows:
-        row[0].setHidden.assert_called_with(False)
+        qtbot.waitUntil(lambda: row[0].setHidden.assert_called_with(False))
```
ZLLentz, line 80: How do these statements work? The encapsuled `assert_called_with` is actually an assert statement. Does it have some undocumented return value?
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with

ZLLentz: Nevermind, I didn't know qtbot would wait until the assertion passed

lightpath/tests/test_widgets.py
tangkong: hah no wonder it never failed 🤦

lightpath/tests/test_gui.py
tangkong: Yea this is some fanciness I picked up while filling out the atef test suite.

lightpath/tests/test_widgets.py
tangkong: Interestingly, locally the `removed` case can seemingly hang forever (10+ seconds, which is effectively forever for a computer brain)

Will investigate further

ZLLentz (approved):
I'll put my approval in because this seems to help

PR #189, opened 2024-04-12
DOC: collect release notes for v1.0.4

ZLLentz:
## Description

Just the one PR, important fix for stability


tangkong (approved):

tangkong:
I've been too lazy to investigate the 3.11 failure... surely it's not a huge problem

ZLLentz:
I'm not super worried about that right now, we can punt. Sometimes these errors go away on their own anyway.

PR #190, opened 2024-08-26
BLD: separate lightpath and lightpath[gui] subpackages

tangkong:
## Description
Separates lightpath base and gui-related dependencies
- pip: `lightpath` and `lightpath[gui]`
- ~conda: `lightpath-base` and `lightpath` (includes gui)~  I could not get conda builds working, we leave this split to the conda feedstock

## Motivation and Context
Lightpath keeps dragging qt dependencies along with it

## How Has This Been Tested?
pip installs locally.  Conda eventually, but let's see if CI takes

This PR

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -39,7 +39,6 @@ test:
     - lightpath
   requires:
     - ipython
-    - matplotlib
     - pytest <7.2.0
     - pytest-qt
 
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -39,7 +39,7 @@ test:
     - lightpath
   requires:
     - ipython
-    - pytest <7.2.0
+    - pytest
     - pytest-qt
 
 about:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -19,7 +19,6 @@ test:
     - lightpath
   requires:
     - ipython
-    - matplotlib
     - pytest <7.2.0
     - pytest-qt
 
@@ -37,6 +36,7 @@ outputs:
         - python >=3.6
         - coloredlogs
         - happi >=1.6.0
+        - networkx <3.3
         - numpy
         - ophyd
         - prettytable
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -27,6 +27,7 @@ outputs:
   - name: lightpath-base
     build:
       noarch: python
+      script: python -m pip install . -vv
     requirements:
       host:
         - python >=3.6
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -49,7 +49,7 @@ outputs:
         - pip
         - setuptools_scm
       run:
-        - lightpath-base
+        - {{ pin_subpackage('lightpath-base', max_pin='x.x') }}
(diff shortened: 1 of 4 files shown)


conda-recipe/meta.yaml
```diff
@@ -43,6 +23,40 @@ test:
...
+  - name: lightpath
+    build:
+      noarch: python
+    requirements:
+      host:
+        - python >=3.6
+        - pip
+        - setuptools_scm
+      run:
+        - lightpath-base
```
ZLLentz, line 52: I think the weird pin invocation from the other recipes is required to make sure lightpath x.x.x depends on lightpath-base at exactly x.x.x

ZLLentz: I don't think this is why the CI is having an issue

tangkong: Adding it did move us forward.  This seems to build fine for me locally so I'll have to abuse GHA a bit here debugging

conda-recipe/meta.yaml
```diff
@@ -13,36 +13,53 @@ source:
...
     - ipython
-    - matplotlib
     - pytest <7.2.0
     - pytest-qt
 
+outputs:
+  - name: lightpath-base
+    build:
+      noarch: python
+      script: python -m pip install . -vv
```
ZLLentz, line 29:
```suggestion
      script: {{ PYTHON }} -m pip install . -vv
```
I'll look for the docs on this

ZLLentz: https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html

tangkong: https://github.com/pcdshub/lightpath/actions/runs/10564732964/job/29267769359#step:12:49
>Attempting to finalize metadata for lightpath-base
Error: Failed to render jinja template in /home/runner/work/lightpath/lightpath/conda-recipe/meta.yaml:
'PYTHON' is undefined

ZLLentz: https://github.com/pcdshub/pcdsdevices/blob/b33b7c4d6a372cb877458f7e446c90d0d083518f/conda-recipe/meta.yaml#L15

tangkong: Maybe this was the warning shot that suggests that something deeper was wrong with the conda build process

ZLLentz: I see now that the recipes that do the *-base thing don't use `{{ PYTHON }}`, my bad for the red herring

tangkong:
adding `--no-deps` managed to remove the clobber warnings I think.  There definitely was some double installations going on, I think what happens is that without `--no-deps`, the pip install installs all the package dependencies, then conda tries to follow the recipe and install them again.  

We still lose the test_env folder though 😢

ZLLentz:
The remaining py311 failures come from pinning pytest<7.2.0, before this version it wasn't marked as noarch so the allowed versions are very pinned down

ZLLentz (dismissed):

tangkong:
Ah we unpinned this a while ago.  This was from an era where pytest-qt wasn't compatible

ZLLentz:
It's all GREEEN

ZLLentz (approved):
This is a great split that will help a lot in various cases!

PR #191, opened 2024-10-15
ENH: add generic rix beamline config

ZLLentz:
## Description

Add a "RIX" beamline config that spans K1 and K2, as a counterpart for the CRIX and QRIX configs.

## Motivation and Context

This is a rix3 hutch-python hotfix verbatim as part of https://jira.slac.stanford.edu/browse/ECS-6333

The point is so that there is a beamline called "RIX" to search for devices on when loading hutch-python.
This beamline ends up having all of the shared RIX devices.

Possibly in the future this would be removed if the rix3 hutch-python was split into crix3 and qrix3 but the way all of this works is unclear to me.

## How Has This Been Tested?

Interactively in rix3 for a year+

Here only.

### DIFF
--- a/lightpath/tests/test_controller.py
+++ b/lightpath/tests/test_controller.py
@@ -5,6 +5,7 @@ import happi
 import pytest
 
 from lightpath import LightController
+from lightpath.config import beamlines
 from lightpath.errors import PathError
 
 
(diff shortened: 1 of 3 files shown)


lightpath/tests/test_controller.py
```diff
@@ -184,6 +185,6 @@ def test_sim_ctrl():
 
     lc = LightController(sim_client)
 
-    assert len(lc.beamlines.keys()) == 9
+    assert len(lc.beamlines.keys()) == len(beamlines)
```
tangkong, line 188: ew did I do this?

tangkong (approved):

lightpath/tests/test_controller.py
ZLLentz: It's not clear to me if you wrote this or just got credited for moving it from one test file to another.
Either way, it's a reminder to all of us to try harder to understand and review the unit tests.

PR #193, opened 2025-03-24
TST: Work to de-flake tests

tangkong:
### Update test_widget logic to make sense, half-removed only applies when upstream of blocker.
- This test was actually incorrect, to the best of my understanding.  Previously we inserted a fully-blocking valve upstream of a partially blocking ipimb, then asserted that the ipimb would be half-removed.  The lightpath logic actually expects this to be removed, and this only ever passed because it wasn't getting updated after the upstream valve was inserted.
- In the lightpath app we actually don't have proper callbacks for updating devices that aren't being inserted/removed.  Was it always this way? 🤷 
- looking at sim lightpath, imager (transmission=0.5, not blocking) after sequence of 
  - insert upstream blocker  -> insert im -> remove upstream blocker 
image
  - insert im, with all other devices removed
image

### test_filter needed to wait for checkbox to update
- after way too many print statements, we were moving on before the value of the "removed" checkbox was ready to reflect the updated value.

## Motivation and Context
py3.12 tests were failing and I want them to pass.

My local testing showed this to be a bit of a race condition.  Lightpath tests were made before I truly understood all the ins and outs of qt testing, so there's probably a couple more places these tests can be tightened up.  Mostly it looks like we're not waiting for the right signals/slots to finish before our assertions.

## How Has This Been Tested?
Interactive testing, and hoping the CI bears it out

This PR

### DIFF
--- a/docs/source/upcoming_release_notes/193-tst_no_flaky.rst
+++ b/docs/source/upcoming_release_notes/193-tst_no_flaky.rst
@@ -1,0 +1,22 @@
+193 tst_no_flaky
+################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Fix a couple tests to improve reproducibility
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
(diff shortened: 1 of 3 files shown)


tangkong:
All my 10x repeat runs have run without failing at this point.

